PR #2, opened 2022-01-28
ADHOC-169: Bash script to add package to projects in a repo

sarahelsaig:
ADHOC-169

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Add package
         id: add_package
         run: |
-          curl https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/scripts/add-github-sourcelink.sh >> add-github-sourcelink.sh
+          curl https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/scripts/add-github-sourcelink.sh > add-github-sourcelink.sh
           bash add-github-sourcelink.sh Microsoft.SourceLink.GitHub
           rm add-github-sourcelink.sh
       - name: Build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,5 +15,11 @@ jobs:
           dotnet-version: 5.0.*
       - name: Install dependencies
         run: dotnet restore -p:NuGetBuild=True
+      - name: Add package
+        id: add_package
+        run: |
+          curl https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/scripts/add-github-sourcelink.sh >> add-github-sourcelink.sh
+          bash add-github-sourcelink.sh Microsoft.SourceLink.GitHub
+          rm add-github-sourcelink.sh
       - name: Build
         run: dotnet build --configuration Release --no-restore  -p:NuGetBuild=True -p:LangVersion=Latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,11 +15,5 @@ jobs:
           dotnet-version: 5.0.*
       - name: Install dependencies
         run: dotnet restore -p:NuGetBuild=True
-      - name: Add package
-        id: add_package
-        run: |
-          curl https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/scripts/add-github-sourcelink.sh > add-github-sourcelink.sh
-          bash add-github-sourcelink.sh Microsoft.SourceLink.GitHub
-          rm add-github-sourcelink.sh
       - name: Build
         run: dotnet build --configuration Release --no-restore  -p:NuGetBuild=True -p:LangVersion=Latest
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -15,5 +15,11 @@ jobs:
           dotnet-version: 5.0.*
       - name: Install dependencies
         run: dotnet restore -p:NuGetBuild=True
+      - name: Add package
```
Piedone: This shouldn't be added to the build pipeline but to the publish one, [before the step I've linked](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/workflows/publish.yml#L40).

scripts/add-github-sourcelink.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+###################################################################################################
+# add-github-sourcelink                                                                           #
```
Piedone: It's nice that it's a generic script (though it needn't be) but then the name and docs need to reflect it.

scripts/add-github-sourcelink.sh
```diff
@@ -0,0 +1,70 @@
...
+# (https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format) for styling.    #
+###################################################################################################
+
+package_name="$1"
+
+function err() {
+    echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*" >&2
+}
+
+function panic() {
```
Piedone: image

.github/workflows/build.yml
```diff
@@ -15,5 +15,11 @@ jobs:
...
+        id: add_package
+        run: |
+          curl https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/scripts/add-github-sourcelink.sh > add-github-sourcelink.sh
```
Piedone: Oh no, there's really no other option? Can't reusable workflows use local scripts (what about maybe putting it into the .github folder too?)?

If it won't work, then plese copy it into a `run` step, what can be copied, but keep an sh file when we want to edit it (and work in a proper editor), with instructions on how to update this step (and instructions in the step on changing the file).

Piedone (changes requested):

.github/workflows/build.yml
sarahelsaig: > Oh no, there's really no other option? Can't reusable workflows use local scripts (what about maybe putting it into the .github folder too?)?

Nope, I've tested putting it into .github and .github/workflows too. You don't get the files in the reusable workflow's directory, so Github itself is probably just directly accessing the workflow file.

> If it won't work, then plese copy it into a run step

Are you sure? I think that's significantly worse in terms of maintainability. What is your concern that drives this extreme move?

Piedone: Apart from the awkwardness of loading part of the pipeline via HTTP separately this also prevents using anything else from the consumer code apart from the latest `dev` version. While this might be what we'll do by default and usually, it's not necessarily the case every time, nor for other users of this projects. Specifying anything else apart from `@dev` [when calling the workflow](https://github.com/Lombiq/Helpful-Libraries/blob/issue/INFRA-92/.github/workflows/publish.yml#L10) would result in confusion, the YML file being the correct version and the script something else.

sarahelsaig: Ok I did that. It nooks genuinely disturbing. Sounds like we are making regular operation much worse just to serve some niche hypothetical.

.github/workflows/publish.yml
```diff
@@ -40,11 +40,69 @@ jobs:
       - name: Add package
         id: add_package
         run: |
-          git clone https://github.com/Lombiq/GitHub-Actions.git /tmp/GitHub-Actions
-          cp /tmp/GitHub-Actions/scripts/add-package.sh .
-          bash add-package.sh Microsoft.SourceLink.GitHub
-          rm -Rf /tmp/GitHub-Actions
-          rm add-package.sh
```
0liver: I found this solution very neat. I didn't really understand, why we can't pull the file from an open repository that we own.

.github/workflows/build.yml
Piedone: For the regular operation this will result in one fewer HTTP call which I'd argue is better :). Editing the script, yes.

.github/workflows/publish.yml
Piedone: See https://github.com/Lombiq/GitHub-Actions/pull/2#discussion_r796065264

.github/workflows/build.yml
0liver: I would argue that an HTTP request is cheap in comparison to copying file content by hand on every change, and more so, remembering it!

I don't see the advantage.

Piedone: I was referring to the regular operation. While I wouldn't care too much about an HTTP request (though it's one additional moving part that can have hiccups any time) this is not my point, see above. This file should change rarely, and when it does, what I've written above will be important.

PR #5, opened 2022-03-26
OSOE-83: Include dot files during 'dotnet pack'

0liver:
OSOE-83

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -139,7 +139,12 @@ jobs:
       - name: Push with dotnet
         run: dotnet nuget push artifacts/*.nupkg --api-key ${{ secrets.API_KEY }} --source ${{ steps.sourceeval.outputs.value }} --skip-duplicate
       - name: Archive artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
         with:
-          name: artifacts
-          path: artifacts          
+          name: NuGet-Package
+          path: artifacts
+      - name: Archive NuGet files
+        uses: actions/upload-artifact@v3
+        with:
+          name: NuGet-Files
+          path: obj/*.nuget.*
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -126,7 +126,7 @@ jobs:
         # * -p:ContinuousIntegrationBuild=true is needed for Deterministic Builds:
         #   https://github.com/clairernovotny/DeterministicBuilds.
         # * -p:DebugSymbols=true and -p:DebugType=portable are needed to generate PDB files.
-        run: dotnet build --configuration Release --no-restore -p:NuGetBuild=true -p:LangVersion=Latest -p:GenerateDocumentationFile=True -p:NoWarn=CS1573%3BCS1591%3BVSTHRD002%3BVSTHRD200 -p:EnableNETAnalyzers=false -p:ContinuousIntegrationBuild=true -p:DebugSymbols=true -p:DebugType=portable --verbosity ${{ inputs.verbosity }}
+        run: dotnet build --configuration Release --no-restore -p:NuGetBuild=true -p:LangVersion=Latest -p:GenerateDocumentationFile=True -p:NoWarn=CS1573%3BCS1591%3BVSTHRD002%3BVSTHRD200 -p:EnableNETAnalyzers=false -p:ContinuousIntegrationBuild=true -p:DebugSymbols=true -p:DebugType=portable --verbosity ${{ inputs.verbosity }} -binaryLogger:build.binlog
       - name: Pack
         # Notes on the configuration apart from what's also for dotnet build:
         # * NoWarn on NU5104 to not have warnings for prerelease dependencies, see:
@@ -148,3 +148,8 @@ jobs:
         with:
           name: NuGet-Files
           path: obj/*.nuget.*
+      - name: Archive log files
+        uses: actions/upload-artifact@v3
+        with:
+          name: Log-Files
+          path: *.binlog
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -131,7 +131,7 @@ jobs:
         #   a non-packagable project.
         # * -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg are needed to generate symbol packages:
         #   https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.
-        run: dotnet pack --output artifacts --configuration Release --no-restore --no-build -p:NuGetBuild=true -p:Version=${{ steps.get_version.outputs.VERSION }} -p:GenerateDocumentationFile=True -p:NoWarn=NU5104 -p:TreatWarningsAsErrors=true -warnaserror -p:WarnOnPackingNonPackableProject=True -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
+        run: dotnet pack --output artifacts --configuration Release --no-restore --no-build -p:NuGetBuild=true -p:Version=${{ steps.get_version.outputs.VERSION }} -p:GenerateDocumentationFile=True -p:NoWarn=NU5104 -p:TreatWarningsAsErrors=true -warnaserror -p:WarnOnPackingNonPackableProject=True -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg -p:NoDefaultExcludes=true
```
0liver: @Piedone The `NoDefaultExcludes` prevents the default exclusion of dot files.

Piedone: Why is this necessary

0liver: Because I have both `.npmrc` and `.stylelintrc` in NE that need to be in the package.

Piedone: This also packs .github folders and .git files.

0liver: Hm, didn't notice that. I thought only those files would be packed that are listed in the project file 🤔

0liver: Where did you see `.git` and `.github` being packed? I can't see any, neither in `Lombiq.Privacy` nor in `NodeJs.Extensions.`

Piedone: This is what the docs say: https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-pack#:~:text=Prevents%20default%20exclusion%20of%20NuGet%20package%20files%20and%20files%20and%20folders%20starting%20with%20a%20dot%2C%20such%20as%20.svn%20and%20.gitignore. However, I tried this out with Helpful Extensions now, and indeed Git files are not packed, so the docs are wrong.

However, I also didn't notice any difference without this config, .stylelintrc is just embedded into the DLL under Resources.

0liver: The point is that I need the dot files as files, not in the DLL ;-)

As for the docs, I'd say they are somewhat misleading. What they are trying to say is that by default no dot files will be packed, even if you specify them to be so. You need to pass that flag to be able to include dot files in your package. Which is what we need.

Piedone: I didn't see  .stylelintrc being included as a content file either. Anyway, it doesn't matter too much, if it works in GitHub Actions without including unwanted files then good, and I see from what I can download that it does work (NE and Privacy).

Piedone (approved):

.github/workflows/publish.yml
0liver: Great then!

.github/workflows/publish.yml
```diff
@@ -2,10 +2,14 @@ name: publish
...
   workflow_call:
-    inputs: 
+    inputs:
       source:
         required: false
         type: string
+      verbosity:
+        required: false
+        type: string
+        default: minimal
```
0liver, line 12: I've added this optional verbosity parameter to more easily test the NuGet build and publish for Lombiq.Privacy. You would use like this:

image

0liver: I'm inclined to think that the same would work for the `source` parameter, which is currently assigned using an extra action (lines 23ff.)

.github/workflows/publish.yml
```diff
@@ -110,32 +114,32 @@ jobs:
...
         #   https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.
-        run: dotnet pack --output artifacts --configuration Release --no-restore --no-build -p:NuGetBuild=true -p:Version=${{ steps.get_version.outputs.VERSION }} -p:GenerateDocumentationFile=True -p:NoWarn=NU5104 -p:TreatWarningsAsErrors=true -warnaserror -p:WarnOnPackingNonPackableProject=True -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
+        run: dotnet pack --output artifacts --configuration Release --no-restore --no-build -p:NuGetBuild=true -p:Version=${{ steps.get_version.outputs.VERSION }} -p:GenerateDocumentationFile=True -p:NoWarn=NU5104 -p:TreatWarningsAsErrors=true -warnaserror -p:WarnOnPackingNonPackableProject=True -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg -p:NoDefaultExcludes=true --verbosity ${{ inputs.verbosity }}
       - name: Push with dotnet
         run: dotnet nuget push artifacts/*.nupkg --api-key ${{ secrets.API_KEY }} --source ${{ steps.sourceeval.outputs.value }} --skip-duplicate
       - name: Archive artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
```
0liver, line 142: I've updated this action to the latest version, which has a lot of nice features that I used for testing. Nothing breaks here.

.github/workflows/publish.yml
Piedone: Add this to [the docs](https://github.com/Lombiq/GitHub-Actions#nuget-publish-workflow) too.

Piedone (changes requested):

Piedone (approved):

PR #7, opened 2022-06-06
OSOE-131: Detect in submodules if there's no PR here

sarahelsaig:
OSOE-131


.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  workflow_call:
+    inputs:
+      repo:
+        required: false
+        type: string
+        default: Lombiq/Open-Source-Orchard-Core-Extensions
+
+env:
+  PULL_REQUEST_TITLE: ${{ github.event.pull_request.title }}
```
0liver, line 12: This variable is now defined both here and in the calling workflow - on purpose? It seems unnecessary. Where should it be?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+          {
+              echo "::error::$Message"
+              exit 1
+          }
+
+          <#
+          .Synopsis
+          Returns $True if any item is coming from the pipe.
+          #>
+          function Any-Object([Parameter(Mandatory=$True, ValueFromPipeline = $True)] $Input)
```
0liver: I think the second formatting with the spaces around the `=` is easier to read, so please make the parameter declaration formatting consistent.

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  parent-pull-request-exists:
+    runs-on: ubuntu-latest
+    name: Parent PR Exists
+    steps:
+      - name: Print PR Title
+        run: echo $Env:PULL_REQUEST_TITLE
+
+      - name: Add Scripts
```
0liver: Let's add some more context for the devs to come:
```suggestion
      - name: Add Commands to Subsequent Workflow Steps
```

.github/workflows/verify.yml
0liver: How did you even edit these scripts? The YAML formatter in VS + Code marks them simply as text. 🤔

EDIT: You can disregard this comment in favor of [this one](https://github.com/Lombiq/GitHub-Actions/pull/7#discussion_r891765989).

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+              return $False
+          }
+          '@
+
+          $ProfileDirectory = [System.IO.Path]::GetDirectoryName($Profile)
+          [System.IO.Directory]::CreateDirectory($ProfileDirectory)
+
+          echo $code >> $Profile
+
+      - name: Current PR Title Includes Issue Code
```
0liver:
```suggestion
      - name: Ensure Current PR Title Includes Issue Code
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          if ($Env:PULL_REQUEST_TITLE -match '^\s*\w+-\d+\s*:') { exit 0 }
+          Set-Failed 'The pull request is not in the right format. Please start with the issue code followed by a colon e.g. "PROJ-123: My PR Title".'
+
+      - name: Search Parent Repository for Matching PR
```
0liver:
```suggestion
      - name: Ensure Parent Repository Contains Matching PR
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+              foreach($item in $Input)
+              {
+                  return $True
+              }
+
+              return $False
+          }
+          '@
+
+          $ProfileDirectory = [System.IO.Path]::GetDirectoryName($Profile)
```
0liver: I couldn't find a definitive guideline, but based on [this longish example](https://docs.microsoft.com/en-us/powershell/scripting/developer/module/how-to-write-a-powershell-script-module?view=powershell-7.2#show-calendar-code-example) and your own convention below I would say that we should use camel-cased variable names everywhere:
```suggestion
          $profileDirectory = [System.IO.Path]::GetDirectoryName($Profile)
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Search Parent Repository for Matching PR
+        run: |
+          $issueCode = $Env:PULL_REQUEST_TITLE -replace '^\s*(\w+-\d+)\s*:.*$', '$1'
+
+          $url = 'https://api.github.com/repos/Lombiq/Open-Source-Orchard-Core-Extensions/pulls?state=open&per_page=100'
+          $titles = curl -H 'Accept: application/vnd.github.v3+json' $url | ConvertFrom-Json | % { $_.title }
+
+          $lookFor = "${issueCode}:"
+          if ($titles | ? { $_ -and $_.Contains($lookFor) } | Any-Object) { exit 0 }
+          Set-Failed "Couldn't find an Open-Source-Orchard-Core-Extensions pull request whose title starts with `"$lookFor`"."
```
0liver: I was asking for the tooling you're using because I remember that we wanted to apply some code styling convention to Powershell scripts by using [VS]Code with some Powershell plugin, and that would discourage (or prohibit) the usage of `% { }` and `? { }` in favor of `Select-Object` and `Where-Object`. Those are arguably longer but they allow for better understanding in the long run.

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Search Parent Repository for Matching PR
+        run: |
+          $issueCode = $Env:PULL_REQUEST_TITLE -replace '^\s*(\w+-\d+)\s*:.*$', '$1'
+
+          $url = 'https://api.github.com/repos/Lombiq/Open-Source-Orchard-Core-Extensions/pulls?state=open&per_page=100'
+          $titles = curl -H 'Accept: application/vnd.github.v3+json' $url | ConvertFrom-Json | % { $_.title }
+
+          $lookFor = "${issueCode}:"
+          if ($titles | ? { $_ -and $_.Contains($lookFor) } | Any-Object) { exit 0 }
```
0liver: Did you invent `Any-Object` for any particular reason? `(array).Count -gt 0` would achieve the same. Or where you explicitly looking to short-circuit the loop?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+          Set-Failed 'The pull request is not in the right format. Please start with the issue code followed by a colon e.g. "PROJ-123: My PR Title".'
+
+      - name: Search Parent Repository for Matching PR
+        run: |
+          $issueCode = $Env:PULL_REQUEST_TITLE -replace '^\s*(\w+-\d+)\s*:.*$', '$1'
+
+          $url = 'https://api.github.com/repos/Lombiq/Open-Source-Orchard-Core-Extensions/pulls?state=open&per_page=100'
+          $titles = curl -H 'Accept: application/vnd.github.v3+json' $url | ConvertFrom-Json | % { $_.title }
+
+          $lookFor = "${issueCode}:"
```
0liver, line 57: Let's move the `$issueCode` closer to its usage. I'd also suggest to use a single variable here, e.g. `$issueCodePrefix`.

.github/workflows/verify.yml
0liver: If we want the PR title to start with the issue code (ignoring whitespace), then let's express that in code:
```suggestion
          if ($titles | ? { $_ -and $_.Trim().StartsWith($lookFor) } | Any-Object) { exit 0 }
```

.github/workflows/verify.yml
0liver: OK, I think we should extract our PS scripts into files and just call them from the workflow [like here](https://www.ntweekly.com/2021/07/28/run-a-powershell-script-on-github-actions/). That way, we can adhere to coding conventions in those scripts and enjoy proper syntax highlighting! 🥳

Even more advantages:
1. Room for proper comments.
2. Shorter workflow files.

0liver:
```suggestion
          Set-Failed "Couldn't find any Open-Source-Orchard-Core-Extensions pull request whose title starts with `"$lookFor`"."
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          if ($Env:PULL_REQUEST_TITLE -match '^\s*\w+-\d+\s*:') { exit 0 }
+          Set-Failed 'The pull request is not in the right format. Please start with the issue code followed by a colon e.g. "PROJ-123: My PR Title".'
+
+      - name: Search Parent Repository for Matching PR
+        run: |
+          $issueCode = $Env:PULL_REQUEST_TITLE -replace '^\s*(\w+-\d+)\s*:.*$', '$1'
+
+          $url = 'https://api.github.com/repos/Lombiq/Open-Source-Orchard-Core-Extensions/pulls?state=open&per_page=100'
+          $titles = curl -H 'Accept: application/vnd.github.v3+json' $url | ConvertFrom-Json | % { $_.title }
```
0liver: I'm wondering if we really need this output here:

image

.github/workflows/verify.yml
```diff
@@ -0,0 +1,74 @@
+name: Verify Submodule Pull Request
```
0liver, line 1: Can you point me to a test run where the verification properly failed?

0liver (changes requested):

.github/workflows/verify.yml
sarahelsaig: It shouldn't change across invocations so only here.

.github/workflows/verify.yml
sarahelsaig: I just edit them as text, hence the inconsistent formatting.

.github/workflows/verify.yml
sarahelsaig: I've linked some in the issue description in Jira.

.github/workflows/verify.yml
sarahelsaig: > and that would discourage (or prohibit) the usage of `% { }` and `? { }` in favor of `Select-Object` and `Where-Object`

If there is any tooling that doesn't let me using `%` and `?` I want nothing to do with it. I've drawn a line in the sand in front of the wagons that I've encircled around this hill that I will die on. 😁 Luckily the linter actually [has a whitelist](https://docs.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingcmdletaliases?view=ps-modules#alias-allowlist). Just like we use `hca` in Orchards we should permit some super common abbreviations like those two in pwsh. I'm willing to use `Get-Childitem` instead of `ls`, but everything has its limits.

I'm also moderately against the [PSUseApprovedVerbs](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2) rule. This isn't an old school text adventure game to only use pre-approved verbs. It takes too long to figure out what verb to shoehorn into a custom function. (Like what should I call `Any-Object`? The best I could come up with is `Measure-NotEmpty` which just rolls off the tongue... directly into the garbage can!) However I think this should be ironed out in a separate issue and with open discussion in Teams.

I've fixed the rest based on the feedback I've got when copying the steps into Code.

> OK, I think we should extract our PS scripts into files and just call them from the workflow [like here](https://www.ntweekly.com/2021/07/28/run-a-powershell-script-on-github-actions/)

Doesn't work, we've already tried this with bash [in a previous issue](https://github.com/Lombiq/GitHub-Actions/pull/2#discussion_r796065264). What you linked works fine for local workflows, but once you try to call into a remote workflow ([like we do here](https://github.com/Lombiq/Orchard-Data-Tables/pull/103/files#diff-b32b798d8ec577b859922a84b90dfe871b19fad404cc6fa7a6fe163f079fbcbfR12)) the situation changes. The content of the remote workflow's repository (this repo) doesn't seem to be accessible. I agree that it would be a good thing if it was possible.

.github/workflows/verify.yml
sarahelsaig: I invented it, becuase I've tried `.Length` and the results were... weird. It returned the first item's string length instead of the collection length so I assumed it's just inconsistent. Count looks good.

.github/workflows/verify.yml
sarahelsaig: Here are some new ones:
- [Ensure Current PR Title Includes Issue Code](https://github.com/Lombiq/Orchard-Data-Tables/actions/runs/2458383201)
- [Ensure Parent Repository Contains Matching PR](https://github.com/Lombiq/Orchard-Data-Tables/runs/6785308647?check_suite_focus=true#step:5:13)

.github/workflows/verify.yml
0liver: You must be kidding me 😢

.github/workflows/verify.yml
```diff
@@ -0,0 +1,59 @@
...
+
+          $profileDirectory = [System.IO.Path]::GetDirectoryName($Profile)
+          [System.IO.Directory]::CreateDirectory($profileDirectory)
+
+          $code | Out-File $Profile
+
+      - name: Ensure Current PR Title Includes Issue Code
+        run: |
+          if ($Env:PULL_REQUEST_TITLE -match '^\s*\w+-\d+\s*:') { exit 0 }
+          Set-Failed 'The pull request is not in the right format. Please start with the issue code followed by a colon e.g. "PROJ-123: My PR Title".'
```
0liver:
```suggestion
          Set-Failed 'The pull request title is not in the expected format. Please start with the issue code followed by a colon and the title, e.g. "PROJ-123: My PR Title".'
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Ensure Parent Repository Contains Matching PR
+        run: |
+          $url = 'https://api.github.com/repos/${{ inputs.repo }}/pulls?state=open&per_page=100'
+          $titles = curl -s -H 'Accept: application/vnd.github.v3+json' $url | ConvertFrom-Json | % { $_.title }
+
+          $issueCode = $Env:PULL_REQUEST_TITLE -replace '^\s*(\w+-\d+)\s*:.*$', '$1'
+          $lookFor = "${issueCode}:"
+          if (($titles | ? { $_ -and $_.Trim().StartsWith($lookFor) }).Count -gt 0) { exit 0 }
+          Set-Failed "Couldn't find any Open-Source-Orchard-Core-Extensions pull request whose title starts with `"$lookFor`"."
```
0liver: Can we maybe help our fellow devs and give them a hint at what they should do now?
```suggestion
          Set-Failed "Couldn't find any Open-Source-Orchard-Core-Extensions pull request whose title starts with `"$lookFor`". Please create one."
```

.github/workflows/verify.yml
0liver: Thanks.

0liver (changes requested):

.github/workflows/verify.yml
0liver: Could [this](https://docs.github.com/en/actions/learn-github-actions/contexts#:~:text=will%20be%20actionscheckout2.-,github.action_path,-string) be of help, maybe? What do you think?

sarahelsaig: I don't see how. Contexts are variables, not files.

0liver: I thought like this maybe:

```yaml
run: |
    ${{ github.action_path }}/myscript.ps1
```

sarahelsaig: Despite the misleading repo name, these are not actions but "[reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)". Each action must have a dedicated repo and they have access to the whole thing as a separate directory on the VM. But that would be a huge overkill for something of this complexity.

I have tried and indeed it doesn't work because `github.action_path` is empty: https://github.com/Lombiq/Orchard-Data-Tables/runs/6794141499?check_suite_focus=true

0liver: Thanks for trying this out. My fingers are itching to dive into this on my own, because it seems unacceptable that extracting scripts into files should not be possible. This should be the way to go, imo.

Anyway, for this issue, let's close this topic.

sarahelsaig: I guess reusing workflows is considered a niche and ill-supported use-case much like using submodules in git. (I'm still searching for a free git client that does submodules well on Linux.)

.github/workflows/verify.yml
```diff
@@ -0,0 +1,59 @@
...
+  PULL_REQUEST_TITLE: ${{ github.event.pull_request.title }}
+
+defaults:
+  run:
+    shell: pwsh
+
+jobs:
+  parent-pull-request-exists:
+    runs-on: ubuntu-latest
+    name: Parent PR Exists
```
0liver: For consistency:
```suggestion
    name: Ensure Parent PR Exists
```

0liver (approved):

.github/workflows/verify.yml
Piedone: Just as a sidenote: While [we could publish our automation code as actions](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows#Ideas-for-the-future), this isn't necessary just for our own stuff, and [we can have any number of actions in a single repo](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows#:~:text=runners%20also%20do.-,DRY%20and%20composition,-%3A%20You%20can%20factor).

PR #8, opened 2022-06-07
OSOE-86: Automate Orchard Core extension manifest version number update during NuGet publish

### LINKED ISSUE #4: Automate Orchard Core extension manifest version number update during NuGet publish (OSOE-86)
When we publish a new NuGet package for OC extensions, we have to maintain `Version` in the extension manifest manually. This is not just necessary for when we use the extension from NuGet, but also when using it as a submodule. Thus, it's not trivial, since we can't just update the manifest for the NuGet publish but also need to do that in the repository.

- Multiple extensions can be in a given repo, each with their own manifests.
- We publish to NuGet by pushing a version tag. However, the version tag should already refer to a commit containing the same version in the manifest... So, perhaps we should move the tag during this process?
- Note that not all projects we publish to NuGet are OC extensions; support for other projects should be kept.
- Related: https://github.com/OrchardCMS/OrchardCore/issues/11335
- We may utilize [Leveraging your CSPROJ meta information](https://docs.orchardcore.net/en/latest/docs/guides/leverage-csproj-meta-info/).

**But!** Perhaps we don' actually need to keep manifest changes in the code; after all, it's mostly us who use the projects as submodules, and even if not, those users have the repo history at their fingertips. So, if the above is too complicated, we could get away with just setting the assembly version (what the build sets) as the manifest version. In this case, we should change all the version numbers in the Manifests to the default (when generated with the CLI template).

dministro:
[OSOE-86](https://lombiq.atlassian.net/browse/OSOE-86)

Fixes GH-4


.github/workflows/publish.yml
```diff
@@ -119,6 +119,43 @@ jobs:
...
+                function Update-Manifest-Version {
+                    param(
+                        [string]
+                        $Manifest,
+                        [string]
+                        $Version
+                    )
+
+                    (Get-Content -Raw $Manifest)`
+                        -replace '(?<head>\[assembly:\s*Module\(([^\]]*Version\W*=\W*"))([^"]*)', ('${head}' + $Version)` |
```
sarahelsaig: What does that backtick do there? I think it's unnecessary

.github/workflows/publish.yml
```diff
@@ -119,6 +119,43 @@ jobs:
...
+                        Out-File $Manifest
```
sarahelsaig: I really dislike the sight of escaping newlines (or any kind of non-printing character).
Also why are you not interpolating `$Version` too?
```suggestion
                    (Get-Content -Raw $Manifest) -replace 
                        '(?<head>\[assembly:\s*Module\(([^\]]*Version\W*=\W*"))([^"]*)', '${head}${Version}' |
                        Out-File $Manifest
```

.github/workflows/publish.yml
```diff
@@ -119,6 +119,43 @@ jobs:
...
+
+                     Write-Output "Version updated in $Manifest to $Version"
+                }
+
+                $manifests = Get-ChildItem $WorkDir -File -Recurse -Filter "Manifest.cs" |
+                               Select-String -List -Pattern 'OrchardCore.Modules.Manifest' | 
+                                 Select-Object -ExpandProperty Path
```
sarahelsaig: I don't think you need to gradually indent a pipe like that.
```suggestion
                $manifests = Get-ChildItem $WorkDir -File -Recurse -Filter "Manifest.cs" |
                    Select-String -List -Pattern 'OrchardCore.Modules.Manifest' | 
                    Select-Object -ExpandProperty Path
```

sarahelsaig (changes requested):

PR #10, opened 2022-06-09
OSOE-86: Automate Orchard Core extension manifest version number update during NuGet publish documentation

dministro:
OSOE-86

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -60,7 +60,7 @@ jobs:
 
 The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the Orchard Core extension manifests and the NuGet packages produced.
 
-There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` or `Theme` definition with the version pushed. Note that, the `Version` property should be present and leave the version number on the default value (0.0.1), this is because we don't actually need to keep manifest version changes in the code.
+There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` or `Theme` definition with the version pushed. Note that the `Version` property should be present and leave the version number on the default value (0.0.1). This is because we don't actually need to keep manifest version changes in the code.
 
 It takes one non-optional secret parameter, `apikey`, the organization API key for pushing to NuGet, and two optional parameters, `source` and `verbosity`. E.g.:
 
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -58,7 +58,9 @@ jobs:
       apikey: ${{ secrets.DEFAULT_NUGET_PUBLISH_API_KEY }}
 ```
 
-The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the NuGet packages produced.
+The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the Orchard Core extension manifests and the NuGet packages produced.
+
+There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` definition with the version pushed.
```
sarahelsaig: I think "no configuration required" is not technically true, because `ModuleAttribute.Version` must be provided in the source code for the regex to update it. `Version` is not a required property and if it's not set, then `ModuleAttribute.DefaultVersionZero` ("0.0") is implied. I'd rather write that the `Version` property must be set to a dummy value, e.g. `0.0.1`. That would also document why it's changed to that for all of our submodules.

sarahelsaig: The "inside the `Module` definition" should be "inside the `Module` or `Theme` definition".

sarahelsaig (changes requested):

Readme.md
```diff
@@ -58,7 +58,9 @@ jobs:
       apikey: ${{ secrets.DEFAULT_NUGET_PUBLISH_API_KEY }}
 ```
 
-The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the NuGet packages produced.
+The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the Orchard Core extension manifests and the NuGet packages produced.
+
+There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` or `Theme` definition with the version pushed. Note that, the `Version` property should be present and leave the version number on the default value (0.0.1), this is because we don't actually need to keep manifest version changes in the code.
```
sarahelsaig: You don't need comma after "Note that", because  it's not meaningful in itself. Also "this is because" is the start of a new sentence.  You could remove the "This is" instead, but I think this is less taxing.
```suggestion
There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` or `Theme` definition with the version pushed. Note that the `Version` property should be present and leave the version number on the default value (0.0.1). This is because we don't actually need to keep manifest version changes in the code.
```

sarahelsaig (changes requested):

Readme.md
dministro: Thank you for your suggestions.

PR #11, opened 2022-06-14
NEST-314: DRY workflows

Piedone:
NEST-314


Docs/Workflows.md
```diff
@@ -0,0 +1,104 @@
...
+- References to projects (`<ProjectReference>` elements) not in the repository won't work, these need to be changed to package references (`<PackageReference>` elements). Make the conditional based on `$(NuGetBuild)`. See the [Helpful Extensions project file](https://github.com/Lombiq/Helpful-Extensions/blob/dev/Lombiq.HelpfulExtensions.csproj) for an example. References to projects in the repository will work and those projects, if configured with the proper metadata, will be published together, with dependencies retained among the packages too.
+- Projects building client-side assets with [Gulp Extensions](https://github.com/Lombiq/Gulp-Extensions) won't work during such builds. Until [we fix this](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/48), you have to commit the *wwwroot* folder to the repository, and add the same conditional to the Gulp and NPM Import elements too ([example](https://github.com/Lombiq/Orchard-Data-Tables/blob/58458b5d6381c71c094cb8d960e12b15a59f62d7/Lombiq.DataTables/Lombiq.DataTables.csproj#L33-L35)).
```
```suggestion
These workflows can be invoked from a step from any other repository's workflow. The utilize [our composite actions](Actions.md).

In addition to the below short explanations and samples, check out the inline documentation of the workflow you want to use, especially its parameters. These examples don't necessarily utilize all parameters.

To add the workflows to a project create a folder in the root of the repository that will call them, e.g. _.github/workflows/build.yml_ and/or _.github/workflows/publish.yml_. Things to keep in mind:

- If you have multiple projects in the repository or if the project you want to build is in a subfolder, then add a solution to the root of the repository that references all projects you want to build.
- References to projects (`<ProjectReference>` elements) not in the repository won't work, these need to be changed to package references (`<PackageReference>` elements). Make the conditional based on `$(NuGetBuild)`. See the [Helpful Extensions project file](https://github.com/Lombiq/Helpful-Extensions/blob/dev/Lombiq.HelpfulExtensions.csproj) for an example. References to projects in the repository will work and those projects, if configured with the proper metadata, will be published together, with dependencies retained among the packages too.
- Projects building client-side assets with [Gulp Extensions](https://github.com/Lombiq/Gulp-Extensions) won't work during such builds. Until [we fix this](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/48), you have to commit the *wwwroot* folder to the repository and add the same conditional to the Gulp and NPM Import elements too ([example](https://github.com/Lombiq/Orchard-Data-Tables/blob/58458b5d6381c71c094cb8d960e12b15a59f62d7/Lombiq.DataTables/Lombiq.DataTables.csproj#L33-L35)).
```

.github/actions/build-dotnet/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    required: false
+    default: .
+    description: Path to the directory where a solution file can be found. Defaults to ".".
+  verbosity:
+    required: false
+    default: quiet
+    description: Verbosity parameter for dotnet build. Defaults to "quiet".
+  enable-code-analysis:
+    required: false
+    default: "true"
```
sarahelsaig, line 15: Is this necessarily `"true"` instead of `true`?

.github/actions/build-dotnet/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    description: Indicates whether to enable static code analysis during build. Defaults to "true".
```
sarahelsaig: The .Net documentation's obsession with "whether" always confused me, becuase I learned in school that if you start the clause with "whether" then it should end with something like "or not".

```suggestion
    description: If set to "true", static code analysis is enabled during the build. Defaults to "true".
```

.github/actions/build-dotnet/action.yml
```diff
@@ -0,0 +1,25 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build and Static Code Analysis
+      shell: pwsh
+      working-directory: ${{ inputs.directory }}
+      run: |
+        ${{ github.action_path }}/build.ps1 -Verbosity ${{ inputs.verbosity }} -EnableCodeAnalysis ${{ inputs.enable-code-analysis }} -Version "1.$Env:GITHUB_RUN_NUMBER.$Env:GITHUB_RUN_ATTEMPT-$Env:GITHUB_RUN_ID"
```
sarahelsaig: Should we use this [splatting](https://get-powershellblog.blogspot.com/2017/07/bye-bye-backtick-natural-line.html#splattingnote) that [Oliver has suggested](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/121#discussion_r888469441) here as well?

.github/actions/build-dotnet/build.ps1
```diff
@@ -0,0 +1,33 @@
+﻿param ($Verbosity, $EnableCodeAnalysis, $Version)
+
+# Notes on build switches that aren't self-explanatory':
```
sarahelsaig:
```suggestion
# Notes on build switches that aren't self-explanatory:
```

.github/actions/build-dotnet/build.ps1
```diff
@@ -0,0 +1,33 @@
...
+# - /p:Retries and /p:RetryDelayMilliseconds are to retry builds if due to random locks it fails the first time.
```
sarahelsaig:
```suggestion
# - /p:Retries and /p:RetryDelayMilliseconds are to retry builds if it fails the first time due to random locks.
```

.github/actions/build-dotnet/build.ps1
```diff
@@ -0,0 +1,33 @@
...
+    "/p:RetryDelayMilliseconds=1000",
+    "-warnAsMessage:MSB3026",
+    "/p:Version=$Version"
+)
+
+Write-Output ".NET version number: $Version"
+
+if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj)
+{
+    Write-Output "Gulp Extensions found. Building it first because it requires it."
```
sarahelsaig: "it requires it" is unclear. What is the first "it"?

.github/actions/setup-azurite/action.yml
```diff
@@ -0,0 +1,20 @@
+﻿name: Set up Azurite
+description: >
+  Sets up the Azurite Azure Blob Storage Emulator
```
sarahelsaig: I haven't seen Auirite referred to as "Azure Blob Storage Emulator" yet. In fact it supports blob, queue and table storage types. If you've accidentally used title case, you don't have to.

```suggestion
  Sets up the Azurite for local Azure Blob Storage emulation
```

.github/actions/setup-dotnet/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    default: 6.0.*
+    description: Version of the .NET SDK to set up. Defaults to 6.0.*.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up .NET
+      env:
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: "true"
+        DOTNET_CLI_TELEMETRY_OPTOUT: "true"
```
sarahelsaig: I think since this is an action, these environment variables won't propagate up to the calling workflow. Is this okay?

.github/actions/setup-sql-server/setup-sql-server.ps1
```diff
@@ -0,0 +1,10 @@
+﻿if ($Env:RUNNER_OS -eq "Windows")
+{
+    choco install sql-server-express --no-progress
+}
+else
+{
+    docker pull mcr.microsoft.com/mssql/server &&
+    docker run --name sql2019 -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password1!' -p 1433:1433 -d 'mcr.microsoft.com/mssql/server:2019-latest' &&
+    docker exec -u 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
+}
```
sarahelsaig:
```suggestion
}

```

.github/actions/setup-sql-server/wait-for-sql-server.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+    if ($i -eq $maxTryCount)
+    {
+        echo "SQL Server couldn't be started."
+        Exit 1
+    }
+
+    echo "SQL Server is not ready. Waiting 1 second."
+    Start-Sleep -s 1
+}
```
sarahelsaig:
```suggestion
}

```

.github/actions/setup-sql-server/setup-sql-server.ps1
```diff
@@ -0,0 +1,10 @@
+﻿if ($Env:RUNNER_OS -eq "Windows")
```
sarahelsaig, line 1: Shouldn't this be an optional switch instead of hard coding by OS? Now that this code is in a reusable action, it feels opinionated.

.github/actions/setup-sql-server/wait-for-sql-server.ps1
```diff
@@ -0,0 +1,30 @@
...
+    {
+        echo "SQL Server is successfully started."
+        Exit 0
+    }
+
+    if ($i -eq $maxTryCount)
+    {
+        echo "SQL Server couldn't be started."
+        Exit 1
+    }
```
sarahelsaig: I think it should be `Write-Error` instead of `echo`

.github/actions/test-dotnet/action.yml
```diff
@@ -0,0 +1,59 @@
+﻿name: Run .NET tests
```
sarahelsaig, line 1: Nice, now I won't have to do this for OSOE-134!

.github/actions/test-dotnet/action.yml
```diff
@@ -0,0 +1,59 @@
...
+          ${{ inputs.build-directory }}/FailureDumps/
+          test.out
+        if-no-files-found: ignore
+
+    - name: Test Report
+      # v8
+      uses: phoenix-actions/test-reporting@1e9b578818d03d6df50ce8e60d024a000bf3ab63
+      if: success() || failure()
+      with:
+        name: Test Results (${{inputs.build-directory}} ${{ matrix.machine-type }})
```
sarahelsaig: Now I think space is not a good separation because the build-directory may contain spaces, however unlikely...
```suggestion
        name: Test Results (${{inputs.build-directory}}|${{ matrix.machine-type }})
```

.github/actions/test-dotnet/action.yml
```diff
@@ -0,0 +1,59 @@
...
+  steps:
+
+    - name: Tests
+      shell: pwsh
+      working-directory: ${{ inputs.build-directory }}
+      run: |
+        ${{ github.action_path }}/run-tests.ps1 -Verbosity ${{ inputs.test-verbosity }}
+
+    # Note that uploading the failure dumps would fail under Windows if the path were too long, regardless of
+    # LongPathsEnabled, see: https://github.com/actions/upload-artifact/issues/309. To get around that we merge all
```
sarahelsaig, line 29: Shouldn't this go into the  PS1 file as documentation?

sarahelsaig (changes requested):

.github/actions/build-dotnet/action.yml
Piedone: Yes. Unlike workflows, the inputs of composite actions can only be strings.

.github/actions/build-dotnet/action.yml
Piedone: It's like this in the .NET world, sorry, whether you like it :D.

.github/actions/build-dotnet/build.ps1
Piedone: Yeah, this is IntelliSense insisting on auto-completed perceived strings or characters in YAML even in comments :|.

.github/actions/build-dotnet/action.yml
sarahelsaig: Wow, how... arbitrary.

.github/actions/build-dotnet/action.yml
sarahelsaig: image

.github/actions/setup-sql-server/setup-sql-server.ps1
Piedone: It surely is, hence the description "Sets up SQL Server with Lombiq-recommended defaults, and if you have a different opinion then too bad because we don't care." Or something. This is, for now, mostly for us. If we want to make them usable for a wider audience then we need to [publish them in the marketplace](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows#Ideas-for-the-future) anyway. There are other SQL Server setup actions BTW with differently opinionated approaches, it doesn't necessarily make sense to publish another one.

I kept the hard-coded password for the same reason.

.github/actions/build-dotnet/action.yml
Piedone: :D

.github/actions/test-dotnet/action.yml
Piedone: I figured it's more appropriate here, since it repeatedly refers to the next step.

.github/actions/setup-dotnet/action.yml
Piedone: Indeed it works like that. BTW I couldn't find any docs on this, only [this comments hints on something similar](https://github.com/actions/runner/pull/1619#issuecomment-1049406644) but not exactly.

It was hilariously tricky but I fixed it.

.github/actions/build-dotnet/action.yml
Piedone: Yeah, cool technique.

.github/actions/test-dotnet/action.yml
sarahelsaig: ok

.github/actions/setup-dotnet/action.yml
```diff
@@ -10,10 +10,13 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Set environment variables
+      shell: pwsh
+      run: |
+        $Env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = "true"
+        $Env:DOTNET_CLI_TELEMETRY_OPTOUT = "true"
+        $Env:SOMETHING_ELSE_JUST_FOR_ME = "hehe"
```
sarahelsaig: image

sarahelsaig (approved):

.github/actions/setup-dotnet/action.yml
Piedone: image

sarahelsaig: image

Piedone: :D

.github/actions/build-dotnet/action.yml
0liver: Especially, when you don't have to quote strings by default (as it seems):

image

.github/actions/build-dotnet/action.yml
0liver: This [answer](https://english.stackexchange.com/a/359727) sums it up quite nicely.

sarahelsaig: Or we could just use "if". I'm pretty sure we are software developers.

Maybe I should make an esolang called DotNet that's just C# but you have to write `whether (isTrue) { doSomething(); }` instead of `if (isTrue) { doSomething(); }`. That should also mess with SEO which is always fun. 😈

PR #13, opened 2022-06-20
OSOE-136: Automate GitHub releases

Piedone:
OSOE-136


.github/workflows/verify.yml
```diff
@@ -1,4 +1,4 @@
-name: Verify Submodule Pull Request
+name: Verify OSOCE Submodule Pull Request
```
sarahelsaig: This change is not needed, the workflow is compatible with any repo. OSOCE is only the default option for `inputs.repo`.

sarahelsaig (changes requested):

PR #14, opened 2022-06-21
OSOE-140: Substitute haya14busa/action-cond with a PS script in the publish workflow

sarahelsaig:
OSOE-140

This has been tested on a temporary branch for Lombiq.AuditTrailExtensions. You can see the successful run [here](https://github.com/Lombiq/Audit-Trail-Extensions/runs/6989302849?check_suite_focus=true) and the successful publish [here](https://www.nuget.org/packages/Lombiq.AuditTrailExtensions#versions-body-tab). I used the `v0.0.0.1-osoe-140.1` tag to avoid publishing an empty pre-release above the current latest version.

### DIFF
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -16,7 +16,7 @@ on:
         required: false
         type: string
         default: 6.0.*
-        description: Version of the .NET SDK to set up. Defaults to "6.0.*".
+        description: Version of the .NET SDK to set up. Defaults to "${{ inputs.dotnet-version.default }}".
       build-directory:
         required: false
         type: string
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -11,32 +11,32 @@ on:
         default: "[\"ubuntu-latest\"]"
         description: >
             Stringified JSON array with the name of the type of machine(s) to run the workflow under, e.g.
-            "[\"ubuntu-latest\"]" or "[\"ubuntu-latest\", \"windows-latest\"]". Defaults to "[\"ubuntu-latest\"]".
+            "[\"ubuntu-latest\"]" or "[\"ubuntu-latest\", \"windows-latest\"]".
       dotnet-version:
         required: false
         type: string
         default: 6.0.*
-        description: Version of the .NET SDK to set up. Defaults to "${{ inputs.dotnet-version.default }}".
+        description: Version of the .NET SDK to set up.
       build-directory:
         required: false
         type: string
         default: .
-        description: Path to the directory where a solution file can be found. Defaults to ".".
+        description: Path to the directory where a solution file can be found.
       build-verbosity:
         required: false
         type: string
         default: quiet
-        description: Verbosity parameter for dotnet build. Defaults to "quiet".
+        description: Verbosity parameter for dotnet build.
       build-enable-code-analysis:
         required: false
         type: string
         default: "true"
-        description: Indicates whether to enable static code analysis during dotnet build. Defaults to "true".
+        description: Indicates whether to enable static code analysis during dotnet build.
       timeout-minutes:
         required: false
         type: number
         default: 360
-        description: Configuration for the timeout-minutes parameter of the workflow. Defaults to 360 (the GitHub default).
+        description: Configuration for the timeout-minutes parameter of the workflow. The 360 is GitHub's default.
 
 jobs:
   # While the below steps seem suitable to DRY with build-and-test-orchard-core, since reusable workflows can't call
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -16,8 +16,7 @@ on:
         required: false
         type: string
         default: 6.0.*
-        description: >
-          Version of the .NET SDK to set up. Defaults to 6.0.*.
+        description: Version of the .NET SDK to set up. Defaults to "6.0.*".
       build-directory:
         required: false
         type: string
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -183,7 +183,21 @@ jobs:
         # * -p:ContinuousIntegrationBuild=true is needed for Deterministic Builds:
         #   https://github.com/clairernovotny/DeterministicBuilds.
         # * -p:DebugSymbols=true and -p:DebugType=portable are needed to generate PDB files.
-        run: dotnet build --configuration Release --no-restore -p:NuGetBuild=true -p:LangVersion=Latest -p:GenerateDocumentationFile=True -p:NoWarn=CS1573%3BCS1591%3BVSTHRD002%3BVSTHRD200 -p:EnableNETAnalyzers=false -p:ContinuousIntegrationBuild=true -p:DebugSymbols=true -p:DebugType=portable --verbosity ${{ inputs.verbosity }}
+        run: |
+          $Arguments = @(
+              "--configuration:Release",
+              "--no-restore",
+              "--verbosity:${{ inputs.verbosity }}",
+              "-p:NuGetBuild=true",
+              "-p:LangVersion=Latest",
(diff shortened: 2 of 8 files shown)


.github/workflows/publish.yml
```diff
@@ -20,16 +20,20 @@ jobs:
...
-        with:
-          cond: ${{ inputs.source == '' }}
-          if_true: "https://api.nuget.org/v3/index.json"
-          if_false: "${{ inputs.source }}"
-
       - name: Print source
-        run: echo NuGet Source:${{ steps.source-value.outputs.value }}
+        shell: pwsh
+        run: |
+          $Source = @'
```
0liver: I thought we agreed on camelCase variable names ❔

.github/workflows/publish.yml
```diff
@@ -20,16 +20,20 @@ jobs:
...
+          ${{ inputs.source }}
+          '@.Trim()
```
0liver: Why do you need 3 lines here instead of a single one?

.github/workflows/publish.yml
```diff
@@ -20,16 +20,20 @@ jobs:
...
-          cond: ${{ inputs.source == '' }}
-          if_true: "https://api.nuget.org/v3/index.json"
-          if_false: "${{ inputs.source }}"
-
       - name: Print source
-        run: echo NuGet Source:${{ steps.source-value.outputs.value }}
+        shell: pwsh
+        run: |
+          $Source = @'
+          ${{ inputs.source }}
```
0liver, line 33: I thought this would work:

```suggestion
          ${{ inputs.source || 'https://api.nuget.org/v3/index.json' }}
```

Have you tried that?

.github/workflows/build-dotnet.yml
```diff
@@ -16,8 +16,7 @@ on:
         required: false
         type: string
         default: 6.0.*
-        description: >
-          Version of the .NET SDK to set up. Defaults to 6.0.*.
+        description: Version of the .NET SDK to set up. Defaults to "6.0.*".
```
0liver: Would it be exaggerated to reuse the default value in the description? Or is this even possible? I'm still not deep enough into GitHub Actions to know. Do you?

```suggestion
        description: Version of the .NET SDK to set up. Defaults to "${{ inputs.dotnet-version.default }}".
```

.github/workflows/publish.yml
```diff
@@ -6,46 +6,59 @@ on:
       source:
         required: false
         type: string
+        default: https://api.nuget.org/v3/index.json
+        description: >
+          The NuGet server URL used by the `dotnet nuget push` command's `--source` argument. 
+          Defaults to "https://api.nuget.org/v3/index.json".
```
0liver: The same would apply here (input parameter default reuse).

.github/workflows/publish.yml
```diff
@@ -6,46 +6,59 @@ on:
...
       verbosity:
         required: false
         type: string
         default: minimal
+        description: The logging verbosity type used by the `dotnet` command.
```
0liver, line 15:
```suggestion
        description: The logging verbosity type used by the `dotnet` command. Defaults to "minimal".
```

.github/workflows/publish.yml
```diff
@@ -170,7 +183,21 @@ jobs:
...
+              "-p:NuGetBuild=true",
+              "-p:LangVersion=Latest",
+              "-p:GenerateDocumentationFile=True",
+              "-p:NoWarn=CS1573%3BCS1591%3BVSTHRD002%3BVSTHRD200",
+              "-p:EnableNETAnalyzers=false",
+              "-p:ContinuousIntegrationBuild=true",
+              "-p:DebugSymbols=true",
+              "-p:DebugType=portable"
+          )
+          dotnet build @Arguments
```
0liver: Sooo nice!

.github/workflows/build-dotnet.yml
0liver: OK, this is probably overkill. 🤭

0liver (changes requested):

.github/workflows/publish.yml
sarahelsaig: I don't remember agreeing to that for pwsh. Do you have a link?

0liver: 👀

.github/workflows/build-dotnet.yml
sarahelsaig: Does it work? Does it look revolting? Also yes. Also why do we even include "Defaults to X" in the description when you can see it right above? As far as I know, there is no GUI for inspecting or importing workflows so there is no benefit.

.github/workflows/publish.yml
0liver: Here's a link to our conversation: https://github.com/Lombiq/GitHub-Actions/pull/7#discussion_r891750813.

But then this seems very convincing: https://poshcode.gitbook.io/powershell-practice-and-style/style-guide/code-layout-and-formatting#:~:text=except%20the%20first.-,powershell%20uses%20pascalcase,-for%20all%20public.

So let's establish one rule and stick to it.

.github/workflows/publish.yml
sarahelsaig: It's a [here-string](https://devblogs.microsoft.com/scripting/powertip-use-here-strings-with-powershell/). This way I won't have to worry if the input contains an apostrophe.

.github/workflows/build-dotnet.yml
0liver: Just because we can, doesn't mean we should.

I agree with dropping the "Defaults to XYZ" everywhere.

.github/workflows/publish.yml
sarahelsaig: Actually if the input is not provided, [it already defaults to](https://github.com/Lombiq/GitHub-Actions/pull/14/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R9) `https://api.nuget.org/v3/index.json`, so there is no real point to this and I'm just going to remove it. If the consumer wisely gave this parameter null or empty string value, then we shouldn't mask the error.

0liver: Agreed. KISS. No offense.

.github/workflows/publish.yml
0liver: Wasn't aware of that one! Thanks.

.github/workflows/publish.yml
sarahelsaig: 😘

.github/workflows/publish.yml
```diff
@@ -6,46 +6,52 @@ on:
...
     steps:
-      # v1.1.0
-      - uses: haya14busa/action-cond@c69871f4d4693e93126e770d7c3588cb1d473462
-        id: source-value
-        with:
-          cond: ${{ inputs.source == '' }}
-          if_true: "https://api.nuget.org/v3/index.json"
-          if_false: "${{ inputs.source }}"
+      - name: Print Source
+        shell: pwsh
```
0liver: This is not needed anymore.

.github/workflows/publish.yml
```diff
@@ -6,46 +6,52 @@ on:
...
+          "SOURCE_URL=$source" >> $Env:GITHUB_ENV
 
-      - name: Get the version
-        id: get-version
+      - name: Get the Version
         run: |
-          VERSION="${GITHUB_REF_NAME//v}"
-          echo VERSION:${VERSION}
-          echo ::set-output name=VERSION::${VERSION}
+          $version = $Env:GITHUB_REF_NAME.Trim().Substring(1)
```
0liver: The following would convey the intent better:
```suggestion
          $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
```

0liver (changes requested):

0liver (approved):

PR #18, opened 2022-06-27
OSOE-149: Adding UI tests parallelism configuration

Piedone:
OSOE-149


.github/actions/test-dotnet/action.yml
```diff
@@ -13,12 +13,32 @@ inputs:
...
+      shell: pwsh
+      run: |
+        if (${{ inputs.ui-test-parallelism }} -eq -1)
+        {
+            Write-Output "The UI test parallelism configuration isn't overridden."
+            Exit 0
+        }
+
+        "Lombiq_Tests_UI__OrchardCoreUITestExecutorConfiguration__MaxParallelTests=${{ inputs.ui-test-parallelism }}" >> $Env:GITHUB_ENV
+        $Env:Lombiq_Tests_UI__OrchardCoreUITestExecutorConfiguration__MaxParallelTests = "${{ inputs.ui-test-parallelism }}"
```
sarahelsaig: It seems to me this line does nothing?

.github/actions/test-dotnet/action.yml
```diff
@@ -13,12 +13,32 @@ inputs:
...
 
-    - name: Tests
+        ${{ github.action_path }}/replace-xunit-maxParallelThreads.ps1 -MaxParallelThreads ${{ inputs.ui-test-parallelism }}
```
sarahelsaig: Why are you using a mixed case file name? I mean this is both camelCase and kebab-case, which means it's
image

Also is there any reason why we aren't naming the ps1 files cmdlet-style? I think in this case that would be `Set-XUnitMaxParallelThreads` if we have [use the approved verbs](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2#:~:text=replaces%20data%20on%20an%20existing%20resource).

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
```diff
@@ -0,0 +1,12 @@
+param ($MaxParallelThreads)
+
+Write-Output "Replacing maxParallelThreads in xunit.runner.json files to $MaxParallelThreads."
+
+$counter = 0
+
+Get-ChildItem "*/**/xunit.runner.json" -Recurse | ForEach {
+     $counter++
+     (Get-Content $_) | ForEach  {$_ -Replace '"maxParallelThreads":\s*([-\d]*)', "`"maxParallelThreads`": $MaxParallelThreads"} | Set-Content $_
+}
```
sarahelsaig: It would be more natural for pwsh to use `| Measure-Object` instead and the use the result's `Count` property.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
```diff
@@ -0,0 +1,12 @@
...
+
+Write-Output "Replaced $counter occurrences."
```
sarahelsaig: Please add a newline at the end.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
```diff
@@ -0,0 +1,12 @@
+param ($MaxParallelThreads)
+
+Write-Output "Replacing maxParallelThreads in xunit.runner.json files to $MaxParallelThreads."
+
+$counter = 0
+
+Get-ChildItem "*/**/xunit.runner.json" -Recurse | ForEach {
```
sarahelsaig: `ForEach` is also an alias and this makes PSScriptAnalyzer angry. I don't think it's a significant improvement over `ForEach-Object` anyway, and it might be confusing with the `foreach` statement considering the case-insensitivity in pwsh. I've already ranted about how `%` is a must for me and that's why I added it to the whitelist, but if you don't want that then I think `ForEach-Object` would be better than going half-measures. What do you think?

sarahelsaig (changes requested):

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
```diff
@@ -0,0 +1,12 @@
...
+     $counter++
+     (Get-Content $_) | ForEach  {$_ -Replace '"maxParallelThreads":\s*([-\d]*)', "`"maxParallelThreads`": $MaxParallelThreads"} | Set-Content $_
```
sarahelsaig: This line is too long. You can add line breaks after the pipe (other binary operators work too) to format the code nicely without having to escape anything.

.github/actions/test-dotnet/action.yml
Piedone: It sets the environment variable for the current step, as opposed to the line above which will just be available in subsequent steps.

.github/actions/test-dotnet/action.yml
Piedone: :D Kebab is for the script name, and I wanted to reference the xUnit config exactly for the rest. Since this is not a cmdlet (nor a function), just a simple script, I wouldn't make its file name look like one's. I renamed the script not to mix cases, though.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
Piedone: `ForEach` already can return a `Count` but that didn't work properly at all. Does I needed to get into the dark art of variable assignment, just for you: https://superuser.com/questions/1456576/powershell-incorrect-array-count-when-piping-foreach-object

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
Piedone: Writing something that can be read out instead of symbol aliases wins in my book. But we can use `ForEach-Object`.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
Piedone: This was actually not too long until using `ForEach-Object` :D.

.github/actions/test-dotnet/action.yml
sarahelsaig: Duh, but where is the variable used in this step? After this line only _replace-xunit-max-parallel-threads.ps1_ is invoked, which in turn doesn't seem to use it anywhere. If you actually use it somewhere, it's not clear so it's worth commenting.

.github/actions/test-dotnet/action.yml
sarahelsaig: What I see is that PSScriptAnalyzer shows a PSUseApprovedVerbs warning on functions regardless whether they are simple functions or cmdlet.
image

If you have to apply the cmdlet style to functions, then why wouldn't you apply it to script file names as well? You invoke either of them the same way and it's trivial to convert one to the other so I'm not sure why should their naming be different.

Alternatively we can throw the cmdlet style naming to the garbage and blacklist the related rules in the PSScriptAnalyzer config. That also works for me, but it should be consistent.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
sarahelsaig: It was 145 characters with `ForEach`. I thought we are doing 120 character limit.

.github/actions/test-dotnet/action.yml
Piedone: Right, it's not needed.

.github/actions/test-dotnet/action.yml
Piedone: But that violation is for a function. We wouldn't call a function like that, and there's no such function here either.

I'm OK with switching over to that style, in Utility Scripts we've actually done that for all scripts (though they're cmdlets anyway). However, this should be consistent in the whole code base and I really don't feel like doing a bunch of renames.

sarahelsaig: If you include the action directory in the PATH then you can easily call a function like that, the difference vanishes.
Anyway, I'm fine with tabling this for later - if you want it done everywhere at once, then I will do the renaming (and the PATH trick) in OSOE-150.

.github/actions/test-dotnet/replace-xunit-maxParallelThreads.ps1
Piedone: > For legibility try to limit lines of code to 120 characters (soft limit) or 150 at most (hard limit, code analyzers will warn about it too), if breaking at 120 would hurt compactness or legibility.

.github/actions/test-dotnet/action.yml
Piedone: Thanks!

PR #19, opened 2022-06-28
OSOE-139: DRY GitHub Actions workflows

sarahelsaig:
OSOE-139

- [Example of `verify` still working as intended.](https://github.com/Lombiq/GitHub-Actions/runs/7093799391?check_suite_focus=true)
- [Example of `publish` still working as intended.](https://github.com/Lombiq/Audit-Trail-Extensions/actions/runs/2579206819)


.gitattributes
```diff
@@ -1,5 +1,8 @@
 * text=auto
 
+# This is to satisfy shellcheck(SC1017).
```
Piedone:
```suggestion
# This is to satisfy shellcheck (SC1017).
```

.github/actions/build-dotnet/action.yml
```diff
@@ -33,6 +33,14 @@ inputs:
       Change this to any other value to alter the cache key, effectively invalidating the latest cache. This is the only
       current way to force "clear" the cache (https://github.community/t/how-to-clear-cache-in-github-actions/129038/5)
       until the associated issue (https://github.com/actions/cache/issues/2) is resolved.
+  publish-version:
+    required: false
+    default: "1.${{ github.run_number }}.${{ github.run_attempt }}-${{ github.run_id }}"
+    description: The value given to `dotnet build`'s `-Version` switch.
```
Piedone, line 39: Can this be markdown?

.github/actions/build-dotnet/action.yml
```diff
@@ -33,6 +33,14 @@ inputs:
       Change this to any other value to alter the cache key, effectively invalidating the latest cache. This is the only
       current way to force "clear" the cache (https://github.community/t/how-to-clear-cache-in-github-actions/129038/5)
       until the associated issue (https://github.com/actions/cache/issues/2) is resolved.
+  publish-version:
+    required: false
+    default: "1.${{ github.run_number }}.${{ github.run_attempt }}-${{ github.run_id }}"
```
Piedone, line 38: Does this really get evaluated as a default?

.github/actions/build-dotnet/action.yml
```diff
@@ -84,10 +92,15 @@ runs:
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
+        $buildSwitches = @'
+        ${{ inputs.dotnet-build-switches }}
+        '@
```
Piedone:
```suggestion
            ${{ inputs.dotnet-build-switches }}
            '@
```

PSScriptAnalyzerSettings.psd1
```diff
@@ -0,0 +1,7 @@
+@{
```
Piedone, line 1: Interesting! Would you document this to https://lombiq.atlassian.net/wiki/spaces/DEV/pages/43384906/PowerShell+and+Windows+Management+Framework, please?

.github/workflows/verify-this-repo.yml
```diff
@@ -0,0 +1,10 @@
+name: Verify Pull Request
```
Piedone, line 1: Why do we need this here?

.github/actions/publish-nuget/action.yml
```diff
@@ -0,0 +1,126 @@
+name: Publish to NuGet
```
Piedone, line 1: Add a note about the new actions to Actions.md.

.github/actions/verify-submodule-pull-request/functions.ps1
```diff
@@ -0,0 +1,10 @@
+<#
+.Synopsis
+Fails the current step with the provided message. Same as "core.setFailed()" from the official toolkit
+(https://github.com/actions/toolkit).
+#>
+function Set-Failed([Parameter(Mandatory = $True)] [string] $Message)
+{
+    Write-Output "::error::$Message"
```
Piedone, line 8: Add a note about these special messages to https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows.

.github/actions/verify-submodule-pull-request/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        echo "Pull Request Title: ${{ inputs.title }}"
+
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+
+        New-Item -ItemType File -Force $Profile | Out-Null
+        Get-Content -Raw ${{ github.action_path }}/functions.ps1 >> $Profile
+
+    - name: Ensure Current PR Title Includes Issue Code
+      shell: pwsh
+      run: Check-Current '${{ inputs.title }}'
```
Piedone, line 30: How come this works without `${{ github.action_path }}`?

.github/actions/publish-nuget/action.yml
```diff
@@ -0,0 +1,126 @@
...
+      shell: pwsh
+      run: dotnet nuget push artifacts/*.nupkg --api-key $Env:API_KEY --source $Env:SOURCE_URL --skip-duplicate
+
+    - name: Publish Artifacts
+      uses: actions/upload-artifact@v3
+      with:
+        name: NuGet-Package
+        path: artifacts
+
+    # This is not in its own action becasuse it's all very specific to this NuGet publishing.
```
Piedone:
```suggestion
    # This is not in its own action because it's all very specific to this NuGet publishing.
```

Yeah, I know :|.

.github/actions/publish-nuget/action.yml
```diff
@@ -0,0 +1,126 @@
+name: Publish to NuGet
+description: Builds all .NET solutions in the given directory with static code analysis.
```
Piedone: :(

.github/actions/publish-nuget/action.yml
```diff
@@ -0,0 +1,126 @@
...
+        ${{ inputs.source }}
+        '@.Trim()
+
+        Write-Output "NuGet Source: $source"
+        "SOURCE_URL=$source" >> $Env:GITHUB_ENV
+        
+        $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+        Write-Output "Target Version: $version"
+        "PUBLISH_VERSION=$version" >> $Env:GITHUB_ENV
+        Write-Output "::set-output name=publish-version::$version"
```
Piedone: Why do we need this both in an env var and output? It would be cleaner to just have the output. Same for `SOURCE_URL`.

.github/actions/publish-nuget/action.yml
```diff
@@ -0,0 +1,126 @@
...
+inputs:
+  source:
+    required: false
+    default: https://api.nuget.org/v3/index.json
+    description: The NuGet server URL used by the `dotnet nuget push` command's `--source` argument. 
+  verbosity:
+    required: false
+    default: minimal
+    description: The logging verbosity type used by the `dotnet` command.
+secrets:
```
Piedone, line 13: VS says this despite it working:
image

Piedone (changes requested):

.github/actions/build-dotnet/action.yml
sarahelsaig: I'm not aware that it's processed by anything so it can be whatever we want. I've started to use backtick in the commit descriptions too, becuase GH renders them as code so I think it's reasonable to use them here too.

.github/actions/build-dotnet/action.yml
sarahelsaig: I have indented `${{ inputs.dotnet-build-switches }}`, becuase every of its lines get trimmed anyway.
However you mustn't indent the `'@` (closing part of the here-string). It must be at the very beginning of the line otherwise it won't work. As far as I understand in YAML that means the same level of indentation as the first line of the snippet. (significant whitespace, spooky!) So that must stay.

.github/workflows/verify-this-repo.yml
sarahelsaig: What do you mean?

.github/actions/verify-submodule-pull-request/action.yml
sarahelsaig: Because I have added `${{ github.action_path }}` to the `$PATH` [here](https://github.com/Lombiq/GitHub-Actions/pull/19/files/360315f11903dd418d7dc0c0105f8d3bbc6faff6#diff-11fc0ebf975145d2e555831bbefb3f686c3bc46455c94e5fe1bb148d7f4785a2R23). 😏

.github/actions/build-dotnet/action.yml
sarahelsaig: Yup, like [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/runs/7103788399?check_suite_focus=true#step:5:60).

image

.github/actions/publish-nuget/action.yml
DAud-IcI: ?

.github/actions/publish-nuget/action.yml
sarahelsaig: You need the output to be pass the value to the "Build" step. I don't think replacing `$Env:PUBLISH_VERSION` with `'${{ steps.setup.outputs.publish-version }}'` for the rest is an improvement of clarity, but it can be done.
As for `SOURCE_URL`, it doesn't have an output counterpart and I prefer to avoid using this `${{ ... }}` thing when possible becuase it makes it harder to locally test a script when it's littered with templating substitutions.

.github/actions/publish-nuget/action.yml
sarahelsaig: I guess the schema used by VS must be out of date. Clearly we need that property there or it will break.

.github/workflows/verify-this-repo.yml
Piedone: Why do we need this workflow verifying PRs in this repo? This isn't an OSOCE submodule in the sense that OSOCE provides any build support (it would be different if the OSOCE build would use the GitHub-Actions submodule's workflows without branch specifiers in YML).

.github/actions/verify-submodule-pull-request/action.yml
Piedone: Ah, cool technique! Added to the wiki.

.github/actions/publish-nuget/action.yml
Piedone: The description is a copy-paste leftover.

.github/actions/publish-nuget/action.yml
Piedone: The Build step already uses `${{ steps.setup.outputs.publish-version }}`. Other steps references `$Env:PUBLISH_VERSION` though. Why can't they also use the output instead?

For `SOURCE_URL`, you can just add an output parameter.

My issue with env vars here is that they're global variables while we have a mechanism to instead explicitly pass values between steps. While using them indeed requires that you locally change them, but for env vars you also have to muddle with your local environment or otherwise change how they're executed locally (and we're talking about one-liner scripts here that we rarely need to change or run locally, I'm not talking about the longer script files; those use parameters anyway).

PSScriptAnalyzerSettings.psd1
sarahelsaig: I've added it to the [PowerShell code conventions](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/43384918/PowerShell+code+conventions).

.github/workflows/verify-this-repo.yml
sarahelsaig: It's a submodule in the sense that it is a submodule :)

My reasoning is that if you change the actions or workflows here, you have to test it on OSOCE too even if it's for a different project. Unfortunately this requires opening a PR and temporarily changing some branch specifiers the way we did with this module issue's OSOCE PR.  Sure, not fully automated, but it's still necessary in my opinion so it's worth reminding.

.github/actions/publish-nuget/action.yml
sarahelsaig: Ok, if you insist. I'm just very uncomfortable with this kind of jumping between different contexts and abstraction levels.
Also environment variables aren't actually global. They are action scoped.

PSScriptAnalyzerSettings.psd1
Piedone: Great! Let's also have https://lombiq.atlassian.net/browse/OSOE-150.

.github/workflows/verify-this-repo.yml
Piedone: Fair enough.

.github/actions/build-dotnet/action.yml
```diff
@@ -41,6 +41,14 @@ inputs:
     required: false
     default: ""
     description: Additional command line switches given to `dotnet build`. You can split these into multiple lines.
+  solution-path:
+    required: false
+    default: "*.sln"
+    description: >
+      The path of the solution file to be built. If you have exactly one .sln file in the current directory then this 
+      can be omitted. Otherwise make sure to specify it to avoid an "MSB1008: Only one project can be specified." error
+      that is thrown when the `*.sln` wildcard results in multiple files. The value is given to Powershell's
```
Piedone:
```suggestion
      that is thrown when the `*.sln` wildcard results in multiple files. The value is given to PowerShell's 
```

Piedone (changes requested):

.github/actions/build-dotnet/action.yml
BenedekFarkas: @DAud-IcI pointed me to this thread from another question, because a few occurrences of this: If this isn't rendered anywhere (as markdown), we're just looking at it as-is, so this means that we have a new way to emphasize something (code included) instead of double quotes. Double quotes work in descriptions and that's what we've been using before, IMHO backtick is a downgrade in terms of readability too.
If it was rendered, then we should use it like this in all descriptions, but we've got inconsistent formatting now.

sarahelsaig: I don't see why backticks are less readable than double quotes, both are surrounding punctuation. Can you explain that?

We are all developers working on GitHub so we all know what backtick means in this context. Double quotes would be a downgrade semantically because they are less specific. We should use backtick to indicate code in all workflow/action descriptions.

Piedone: Double quotes in these descriptions only work well because YAML is liberal with how it handles strings (or should I say, inconsistent), since wrapping the full description in quotes will be understood as a string, but you can also omit the quotes, but if you add quotes in the middle of the description then the whole description will still be understood as one string... My point is, that backticks are fine for code, and if we can get away without surrounding quotes (like on line 38 above) then we should do that. Otherwise, use quotes in the descriptions according to English grammar.

PR #21, opened 2022-06-30
 OSOE-110: Fix `verify` action that fails if the PR title has apostrophe in it

sarahelsaig:
OSOE-110

- [Example of the problem.](https://github.com/Lombiq/.NET-Analyzers/runs/7125011621?check_suite_focus=true)
- [Info on using `env` to set a variable just for a specific step.](https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#using-variables-in-your-workflows)


.github/actions/build-dotnet/Build-DotNetSolution.ps1
```diff
@@ -50,6 +50,12 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
     Write-Output "::endgroup::"
 }
 
+# This prepares the solution with the Lombiq.Analyzers files. The output and exit code are dismissed because they will
+# be in error if there is a project without the LombiqNetAnalyzers target. Then there is nothing to do, and the target 
+# will still run on the projects that have it.
+dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $Solution | Out-Null
+bash -c 'true'
```
0liver: What do we need that for?

.github/actions/publish-nuget/action.yml
```diff
@@ -92,19 +102,20 @@ runs:
       # * -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg are needed to generate symbol packages:
       #   https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.
       run: |
-        $noWarn = @('NU5104') + @'
+        $noWarn = @('NU5104', 'NU5100') + @'
```
0liver: It would be beneficial to include a comment that explains why NU5104 and NU5100 are being ignored so the occasional reader doesn't have to look those up.

0liver: Order error codes alphabetically just for the fun of it.

.github/actions/publish-nuget/action.yml
```diff
@@ -92,19 +102,20 @@ runs:
...
             "--no-restore",
             "--no-build",
-            "--output:artifacts",
+            "--output:" + (Join-Path $PWD artifacts),
             "--verbosity:${{ inputs.verbosity }}",
             "-p:NuGetBuild=true",
             "-p:Version=${{ steps.setup.outputs.publish-version }}",
+            "-p:NuspecProperties=version=${{ steps.setup.outputs.publish-version }}",
             "-p:GenerateDocumentationFile=True",
             "-p:NoWarn=$($noWarn -join '%3B')",
```
0liver, line 124:
```suggestion
            "-p:NoWarn=$($noWarn -join '%3B')", # Escape the semicolon to prevent the statement from terminating here.
```
Or something along the lines.

.github/actions/publish-nuget/New-NugetPackages.ps1
```diff
@@ -0,0 +1,18 @@
+param([array] $Arguments)
```
0liver: Add a short description of what this script does.

.github/actions/verify-submodule-pull-request/action.yml
```diff
@@ -25,10 +25,16 @@ runs:
         New-Item -ItemType File -Force $Profile | Out-Null
         Get-Content -Raw ${{ github.action_path }}/functions.ps1 >> $Profile
 
+    # We are using env for the title input in these steps because it realistically could contain 
+    # quotes apostrophes or even both. This way those don't need to be escaped.
```
0liver: I don't know if we'll find tooling for that, but let's wrap comments at 120 chars here, too:
```suggestion
    # We are using env for the title input in these steps because it realistically could contain quotes, apostrophes or
    # even both. This way those don't need to be escaped.
```

.github/workflows/test-analysis-failure.yml
```diff
@@ -87,7 +87,7 @@ jobs:
       uses: Lombiq/GitHub-Actions/.github/actions/enable-corepack@dev
 
     - name: Build and Static Code Analysis
-      uses: Lombiq/GitHub-Actions/.github/actions/build-dotnet@dev
+      uses: Lombiq/GitHub-Actions/.github/actions/build-dotnet@issue/OSOE-110
```
0liver, line 90: I don't see any issue-specific changes in the `build-dotnet/action.yml` - shouldn't the issue branch specifier be unnecessary in that case? Or am I missing something? If not, then the dependent invocations can also just point to `dev`.

.github/actions/publish-nuget/action.yml
```diff
@@ -92,19 +102,20 @@ runs:
       # * -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg are needed to generate symbol packages:
       #   https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.
       run: |
-        $noWarn = @('NU5104') + @'
+        $noWarn = @('NU5104', 'NU5100') + @'
         ${{ inputs.dotnet-pack-ignore-warning }}
         '@.Split() | ? { $_ }
```
0liver, line 111: If we used `[string]::IsNullOrEmpty($_)` in [Build-DotNetSolution.ps1](https://github.com/Lombiq/GitHub-Actions/pull/21/files#diff-4cf358b6f80e00ce9cd739a04c94dfad647b09ca5d6a524757944a1c611d6848R13), we should do the same here, don't you think? Or the other way round. But consistently.

Sorry for picking on this unchanged line 😏

.github/actions/publish-nuget/action.yml
```diff
@@ -92,19 +102,20 @@ runs:
...
 
-        $arguments = @(
+        New-NugetPackages @(
```
0liver: Why exactly do have to switch from `pack`ing the solution to `pack`ing individual projects?

0liver (changes requested):

.github/actions/build-dotnet/Build-DotNetSolution.ps1
sarahelsaig: It's written in the comment above. To dismiss the exit code for the preceding command. For some reason doing the same thing with pwsh doesn't work. Moved it to the preceding line with `||` for improved clarity. (it's more clear for bash people at least)

.github/actions/publish-nuget/action.yml
sarahelsaig: I will buy you dictionary for x-mas. :P

sarahelsaig: There is already a comment for NU5104 above the step. Added one for NU5100 too

.github/actions/publish-nuget/action.yml
sarahelsaig: If you look through the scripts in yml, there are no inline comments anywhere. I'm not trusting that the lines won't get wrapped, se let's not start now. Also I don't think this needs to be explained here specifically when '%3B' is used all over the repo.

.github/workflows/test-analysis-failure.yml
sarahelsaig: _Build-DotNetSolution.ps1_ was changed which is called by _build-dotnet/action.yml_

.github/actions/publish-nuget/action.yml
sarahelsaig: Ok, the other way around.

.github/actions/publish-nuget/action.yml
sarahelsaig: You can only specify a custom _nuspec_ file when you pack an individual project.

.github/actions/publish-nuget/action.yml
0liver: It wasn't obvious for me. Obviously, once you put on your escape goggles 🥽 you can see semicolons everywhere.

I understand your concern with the inline comment. OK then.

.github/actions/publish-nuget/New-NugetPackages.ps1
0liver: Maybe it's just me getting acquainted with your code slower than I'd like, but I think a two-liner about the intention of a script is a good habit. Saves a lot of time when you come here for the first time, or the first time after a long time, and need to get a grasp of the whole system. It slows me down when I have to analyze the code in detail to understand what it does, when I really don't need to know the details (at the moment).

.github/actions/publish-nuget/action.yml
0liver: That's valuable information, probably suited for `New-NuGetPackages`.

.github/actions/build-dotnet/Build-DotNetSolution.ps1
0liver: I'm surprised that you could run a command with an error exit code in one line (`dotnet msbuild`) and then simply "override" its exit code. I would expect the failing command to error out the script right away.

Also, despite the fact that "dismiss" might be the correct way of expressing this, I think reading "discard" would make it even clearer what's happening here. Although the `|| bash -c 'true'` is already helpful.

.github/actions/publish-nuget/New-NugetPackage.ps1
```diff
@@ -0,0 +1,32 @@
+
```
0liver: Remove empty line. Save a 🌴 ! 😉

.github/actions/publish-nuget/New-NugetPackages.ps1
0liver: Thank you!

.github/actions/verify-submodule-pull-request/action.yml
0liver: That comma I snuck in hid itself from view 😁

.github/workflows/test-analysis-failure.yml
0liver: Very subtle!

.github/actions/publish-nuget/New-NugetPackage.ps1
```diff
@@ -0,0 +1,32 @@
...
+<#
+.SYNOPSIS
+    Creates a NuGet package from each project in the sln file in the current directory.
+.DESCRIPTION
+    Uses "dotnet sln list" to get all projects in the current directory. This means the current directory must have 
+    exactly one sln file in it. Then calls "dotnet pack" for each csproj file with the provided arguments. If there is
+    a nuspec file in the project's directory too, then it is used to generate the package description instead of the
```
0liver:
```suggestion
    a nuspec file in the project's directory, then that is used to generate the package description instead of the 
```

.github/actions/publish-nuget/New-NugetPackage.ps1
```diff
@@ -0,0 +1,32 @@
+
+<#
+.SYNOPSIS
```
0liver: I'd indent this [like here](url).

.github/actions/verify-submodule-pull-request/action.yml
```diff
@@ -25,10 +25,16 @@ runs:
         New-Item -ItemType File -Force $Profile | Out-Null
         Get-Content -Raw ${{ github.action_path }}/functions.ps1 >> $Profile
 
+    # We are using env for the title input in these steps because it realistically could contain quotes apostrophes or
```
0liver:
```suggestion
    # We are using env for the title input in these steps because it realistically could contain quotes, apostrophes, or
```

.github/actions/publish-nuget/New-NugetPackages.ps1
sarahelsaig: I'm confused about this two-liner idea. I haven't seen such a comment in any of our PowerShell scripts. Also if you are going to add documentation it seems like a waste to add the kind you can only see if you inspect the script file.

.github/actions/verify-submodule-pull-request/action.yml
0liver: [Serial comma](https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/) or not? I'm in favor.

0liver (changes requested):

.github/actions/publish-nuget/New-NugetPackage.ps1
sarahelsaig: Your url is broken. Also this is the result of the `comment-help` snippet in VS Code ([see](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/43384918/PowerShell+code+conventions#:~:text=just%20start%20typing%20comment-help%20a%20line%20above%20the%20function%20and%20select%20it%20in%20the%20intellisense%20suggestions.)) from the official PowerShell extension so they probably know what they are doing.

.github/actions/verify-submodule-pull-request/action.yml
sarahelsaig: I don't think it's applicable here, because "or even both" doesn't create that kind of ambiguity. Also it confuses any comma counting tooling ([we have that for markdown for example](https://www.npmjs.com/package/textlint-rule-max-comma)) that expect common comma usage.
I'm in favor of oxford comma only when it helps, but not as a general policy.

.github/actions/publish-nuget/New-NugetPackage.ps1
0liver: Funny, because [the example](https://docs.microsoft.com/en-us/powershell/scripting/developer/help/examples-of-comment-based-help?view=powershell-7.2#example-2-comment-based-help-for-a-script) I was referring to is also from the MS PS team. But ➕ 1️⃣ for not adjusting auto-generated snippets.

.github/actions/build-dotnet/Build-DotNetSolution.ps1
```diff
@@ -50,6 +48,11 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
     Write-Output "::endgroup::"
 }
 
+# This prepares the solution with the Lombiq.Analyzers files. The output and exit code are discarded because they will
+# be in error if there is a project without the LombiqNetAnalyzers target. Then there is nothing to do, and the target 
+# will still run on the projects that have it.
+dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $Solution | Out-Null || bash -c 'true'
```
0liver, line 54: So here's my confusion: In the [Build.props](https://github.com/Lombiq/.NET-Analyzers/pull/57#discussion_r935460867) you write that we need to run `dotnet msbuild -target:LombiqNetAnalyzers` but here we also call the `Restore` target. Why?

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -0,0 +1,32 @@
...
+    a nuspec file in the project's directory too, then it is used to generate the package description instead of the 
+    regular auto-generation.
+.EXAMPLE
+    New-NugetPackage @("--configuration:Release", "--warnaserror")
+    Calls "dotnet pack project.csproj --configuration:Release --warnaserror" on each project.
+#>
+
+param([array] $Arguments)
+
+# We go through the projects individually, because the "-p:NuspecFile=" parameter can't be passed to a solution.
```
0liver: Since this is the main reason for having this script in the first place, please add this information to the DESCRIPTION.

0liver (changes requested):

.github/actions/build-dotnet/Build-DotNetSolution.ps1
sarahelsaig: > [If you are using NuGet, then the argument should be "-target:Restore;LombiqNetAnalyzers" instead (include the quotes) to restore all package dependencies first.](https://github.com/Lombiq/.NET-Analyzers/blob/101ccada871b31c8896d87f3c296e4ca87e7ddc3/Lombiq.Analyzers/Build.props#L78-L79)

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: I will do it because I'm sick of this issue and want to get it over with already. But I also disagree. While this is the technical reason to justify a script, from an invoker perspective it's only an implementation detail.

0liver (approved):

PR #30, opened 2022-08-05
OSOE-165: Fixing NuGet publishing for repositories without an sln file

Piedone:
OSOE-165

### DIFF
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -16,14 +16,7 @@
 
 param([array] $Arguments)
 
-if (Test-Path *.sln)
-{
-    $projects = (dotnet sln list | Select-Object -Skip 2 | Get-Item)
-}
-else
-{
-    $projects = Get-ChildItem *.csproj
-}
+$projects = (Test-Path *.sln) ? (dotnet sln list | Select-Object -Skip 2 | Get-Item) : (Get-ChildItem *.csproj)
 
 foreach ($project in $projects)
 {
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -16,13 +16,13 @@
 
 param([array] $Arguments)
 
-if (!(Get-ChildItem *.sln))
+if (Test-Path *.sln)
 {
-    $projects = Get-ChildItem *.csproj
+    $projects = (dotnet sln list | Select-Object -Skip 2 | Get-Item)
 }
 else
 {
-    $projects = (dotnet sln list | Select-Object -Skip 2 | Get-Item)
+    $projects = Get-ChildItem *.csproj
 }
 
 foreach ($project in $projects)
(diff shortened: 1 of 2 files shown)


.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,7 +16,16 @@
 
 param([array] $Arguments)
 
-foreach ($project in (dotnet sln list | Select-Object -Skip 2 | Get-Item))
+if (!(Get-ChildItem *.sln))
```
sarahelsaig: Let's stick with `-not` for neqation instead of the undocumented exclamation mark. However if you swap the if-else then you don't need any negation in the first place and save an extra pair of parens too.

Also `Test-Path` would be more expressive here than `Get-ChildItem`.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,7 +16,16 @@
...
+{
+    $projects = Get-ChildItem *.csproj
+}
+else
+{
+    $projects = (dotnet sln list | Select-Object -Skip 2 | Get-Item)
+}
```
sarahelsaig: In PowerShell statements return a value too. This is so beautiful it brings tears to my eyes.
So you can do
```pwsh
$projects = if ( ... )
{
}
else
{
}
```
instead of assigning to the variable in each branch.

sarahelsaig (changes requested):

.github/actions/publish-nuget/New-NuGetPackage.ps1
0liver: 😂 or 😢 ?

.github/actions/publish-nuget/New-NuGetPackage.ps1
Piedone: Isn't the first one something the analyzers should catch?

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: 🥲

Piedone: Seems like poor man's ternary operator to me! Fortunately, PS 7 has actual ternary operators. I briefly contemplated keeping Windows PowerShell compatibility.

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: I agree, it should. [I've opened an issue.](https://github.com/PowerShell/PSScriptAnalyzer/issues/1826)

Piedone: Thank you.

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: It's more like a rich man's ternary operator because you can have multiple expressions in each arm so it's more powerful than a ternary. But yes, please use ternary then. This repo is [already not compatible](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/setup-sql-server/Initialize-SqlServer.ps1#L7-L9) with Windows PowerShell.

Piedone: Indeed, and it's lavish with all the characters. Almost like showing off! :D

sarahelsaig: image

Piedone: :D

PR #42, opened 2022-08-17
OSOE-172: Check if all packages are consolidated in a solution

### LINKED ISSUE #31: Check if all packages are consolidated in a solution (OSOE-172)
In the .NET build workflows, add an optional (enabled by default) check that fails if NuGet packages are not consolidated across the given solution.

You may use [`dotnet-consolidate`](https://github.com/olsh/dotnet-consolidate) for this but do check if there's a built-in .NET CLI feature for this.

Piedone:
OSOE-172

Fixes #31.

### DIFF
--- a/.github/actions/verify-dotnet-consolidation/action.yml
+++ b/.github/actions/verify-dotnet-consolidation/action.yml
@@ -12,7 +12,7 @@ inputs:
     required: false
     default: "*.sln"
     description: >
-      The path of the solution file to be built. The value is given to PowerShell's `Get-ChildItem` cmdlet as-is, so
+      The path of the solution file to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet as-is, so
       grepping is still possible but a name with spaces must be escaped separately.
 
 runs:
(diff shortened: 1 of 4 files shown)


.github/actions/verify-dotnet-consolidation/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  directory:
+    required: false
+    default: .
+    description: Path to the directory where a solution file can be found.
+  solution-path:
+    required: false
+    default: "*.sln"
+    description: >
+      The path of the solution file to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet as-is, so
+      grepping is still possible but a name with spaces must be escaped separately.
```
dministro, line 16:
```suggestion
      The path of the solution file is to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet as-is, so
      grepping is still possible but a name with spaces must be escaped separately.
```
`Grammarly` not me

dministro (changes requested):

.github/actions/verify-dotnet-consolidation/action.yml
Piedone: Grammarly is wrong in this case :). If the sentence was alone, it would make sense, but not here. If this would be out of the context of the action parameter, the sentence would be: "`solution-path` is a parameter for the path of the solution file to be processed." This is a truncated sentence in the comment.

dministro (approved):

PR #43, opened 2022-08-22
OCC-29: Use steps.setup.outputs.publish-version

sarahelsaig:
instead of `github.ref` because `$Env:GITHUB_REF_NAME` is not used in every scenario any more.


DemeSzabolcs:
Closed this one, since because of the naming it couldn't verify the parent PR.  Because of "\`example\`" it doesn't terminated the `string`. And it used the same title even after renaming it.

I opened a new one: https://github.com/Lombiq/GitHub-Actions/pull/44

PR #46, opened 2022-08-31
NEST-316: Re-implement Azure Deploy build with GitHub Actions

DemeSzabolcs:
NEST-316

### DIFF
--- a/.github/actions/build-dotnet/Build-DotNetSolution.ps1
+++ b/.github/actions/build-dotnet/Build-DotNetSolution.ps1
@@ -1,6 +1,5 @@
 param (
-    [string] $Solution,
-    [string] $WebProject,
+    [string] $SolutionOrWebProject,
     [string] $Verbosity,
     [string] $EnableCodeAnalysis,
     [string] $Version,
@@ -34,8 +33,6 @@ $buildSwitches = ConvertTo-Array @"
     $Switches
 "@
 
-$applicationToBeBuilt =  if ([string]::IsNullOrEmpty($WebProject)) {$Solution} else {$WebProject}
-
 [array] $expectedErrorCodes = ConvertTo-Array $ExpectedCodeAnalysisErrors | % { $_.Split(':')[0] } | Sort-Object
 $noErrors = $expectedErrorCodes.Count -eq 0
 
@@ -51,18 +48,18 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
     Write-Output "::endgroup::"
 }
 
-# This prepares the solution with the Lombiq.Analyzers files. The output and exit code are discarded because they will
+# This prepares the solution or web project with the Lombiq.Analyzers files. The output and exit code are discarded because they will
 # be in error if there is a project without the LombiqNetAnalyzers target. Then there is nothing to do, and the target 
 # will still run on the projects that have it.
-dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $Solution | Out-Null || bash -c 'true'
+dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $SolutionOrWebProject | Out-Null || bash -c 'true'
 
-Write-Output "Building solution/web project with ``dotnet build $applicationToBeBuilt $($buildSwitches -join " ")``."
+Write-Output "Building solution or web app with ``dotnet build $SolutionOrWebProject $($buildSwitches -join " ")``."
 
 $errorLines = New-Object "System.Collections.Generic.List[string]"
 $errorCodes = New-Object "System.Collections.Generic.List[string]"
 
 $errorFormat = '^(.*)\((\d+),(\d+)\): error (.*)'
-dotnet build $Solution @buildSwitches 2>&1 | % {
+dotnet build $SolutionOrWebProject @buildSwitches 2>&1 | % {
     if ($_ -notmatch $errorFormat) { return $_ }
 
     ($null, $file, $line, $column, $message) = [regex]::Match($_, $errorFormat, 'Compiled').Groups.Value
@@ -110,7 +107,7 @@ if ($expectedErrorCodes)
         exit 1
     }
 
-    Write-Output "Verification complete, the solution/web project only has the expected errors!"
+    Write-Output "Verification complete, the solution or web project only has the expected errors!"
     exit 0
 }
 
--- a/.github/actions/build-dotnet/Build-DotNetSolution.ps1
+++ b/.github/actions/build-dotnet/Build-DotNetSolution.ps1
@@ -1,5 +1,5 @@
 param (
-    [string] $SolutionOrWebProject,
+    [string] $SolutionOrProject,
     [string] $Verbosity,
     [string] $EnableCodeAnalysis,
     [string] $Version,
@@ -48,18 +48,18 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
     Write-Output "::endgroup::"
 }
 
-# This prepares the solution or web project with the Lombiq.Analyzers files. The output and exit code are discarded because they will
+# This prepares the solution or project with the Lombiq.Analyzers files. The output and exit code are discarded because they will
 # be in error if there is a project without the LombiqNetAnalyzers target. Then there is nothing to do, and the target 
 # will still run on the projects that have it.
-dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $SolutionOrWebProject | Out-Null || bash -c 'true'
+dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $SolutionOrProject | Out-Null || bash -c 'true'
 
-Write-Output "Building solution or web app with ``dotnet build $SolutionOrWebProject $($buildSwitches -join " ")``."
+Write-Output "Building solution or project with ``dotnet build $SolutionOrProject $($buildSwitches -join " ")``."
 
 $errorLines = New-Object "System.Collections.Generic.List[string]"
 $errorCodes = New-Object "System.Collections.Generic.List[string]"
 
 $errorFormat = '^(.*)\((\d+),(\d+)\): error (.*)'
-dotnet build $SolutionOrWebProject @buildSwitches 2>&1 | % {
+dotnet build $SolutionOrProject @buildSwitches 2>&1 | % {
     if ($_ -notmatch $errorFormat) { return $_ }
 
     ($null, $file, $line, $column, $message) = [regex]::Match($_, $errorFormat, 'Compiled').Groups.Value
@@ -107,7 +107,7 @@ if ($expectedErrorCodes)
         exit 1
     }
 
-    Write-Output "Verification complete, the solution or web project only has the expected errors!"
+    Write-Output "Verification complete, the solution or project only has the expected errors!"
     exit 0
 }
 
--- a/.github/actions/build-dotnet/Build-DotNetSolution.ps1
+++ b/.github/actions/build-dotnet/Build-DotNetSolution.ps1
@@ -1,5 +1,6 @@
-﻿param (
+param (
     [string] $Solution,
+    [string] $WebProject,
     [string] $Verbosity,
     [string] $EnableCodeAnalysis,
     [string] $Version,
@@ -33,6 +34,8 @@ $buildSwitches = ConvertTo-Array @"
     $Switches
 "@
 
+$applicationToBeBuilt =  if ([string]::IsNullOrEmpty($WebProject)) {$Solution} else {$WebProject}
+
 [array] $expectedErrorCodes = ConvertTo-Array $ExpectedCodeAnalysisErrors | % { $_.Split(':')[0] } | Sort-Object
 $noErrors = $expectedErrorCodes.Count -eq 0
 
@@ -53,7 +56,7 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
 # will still run on the projects that have it.
 dotnet msbuild '-target:Restore;LombiqNetAnalyzers' $Solution | Out-Null || bash -c 'true'
 
-Write-Output "Building solution with ``dotnet build $Solution $($buildSwitches -join " ")``."
+Write-Output "Building solution/web project with ``dotnet build $applicationToBeBuilt $($buildSwitches -join " ")``."
 
 $errorLines = New-Object "System.Collections.Generic.List[string]"
 $errorCodes = New-Object "System.Collections.Generic.List[string]"
@@ -107,7 +110,7 @@ if ($expectedErrorCodes)
         exit 1
     }
 
-    Write-Output "Verification complete, the solution only has the expected errors!"
+    Write-Output "Verification complete, the solution/web project only has the expected errors!"
     exit 0
 }
 
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -42,21 +42,14 @@ inputs:
     required: false
     default: ""
     description: Additional command line switches given to `dotnet build`. You can split these into multiple lines.
-  solution-path:
+  solution-or-web-project-path:
     required: false
     default: "*.sln"
     description: >
-      The path of the solution file to be built. If you have exactly one .sln file in the current directory then this
+      The path of the solution file or web app to be built. If you have exactly one .sln file in the current directory then this
       can be omitted. Otherwise make sure to specify it to avoid an "MSB1008: Only one project can be specified." error
       that is thrown when the `*.sln` wildcard results in multiple files. The value is given to PowerShell's
       `Get-ChildItem` cmdlet as-is, so grepping is still possible but a name with spaces must be escaped separately.
-  web-project-path:
-    required: false
-    default: ""
-    description: >
-      The path of the web project file to be built. The value is given to PowerShell's `Get-ChildItem` cmdlet
-      as-is, so grepping is still possible but a name with spaces must be escaped separately. If specified this will be
-      built instead of the solution.
   expected-code-analysis-errors:
     required: false
     default: ""
@@ -129,8 +122,7 @@ runs:
         '@
 
         $switches = @{
-            Solution = (Get-ChildItem ${{ inputs.solution-path }}).FullName
-            WebProject = (Get-ChildItem ${{ inputs.web-project-path }}).FullName
+            Solution = (Get-ChildItem ${{ inputs.solution-or-web-project-path }}).FullName
             Verbosity = "${{ inputs.verbosity }}"
             EnableCodeAnalysis = "${{ inputs.enable-code-analysis }}"
             Version = "${{ inputs.publish-version }}"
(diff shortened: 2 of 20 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
+name: Deploy
```
Piedone: Rather call this and the workflows in DotNest too, "Deploy to Azure App Service", and similarly with the file. This will also fix IntelliSense errors in VS that are displayed in the editor for some reason if the file name is just "deploy".

Docs/Workflows.md
```diff
@@ -98,4 +98,28 @@ jobs:
...
+    name: Deploy
+    uses: Lombiq/GitHub-Actions/.github/workflows/deploy.yml@dev
+    with:
+      machine-types: "[\"ubuntu-latest\"]"
+      timeout-minutes: 60
+      app-name: AppName
+      resource-group-name: ResourceGroupName
+      slot-name: Staging
+      url: https://www.myapp.com
+    secrets: inherit
```
Piedone: This passes all the organization secrets to the workflow too. That's a security risk, we only want to pass those secrets that are actually required.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+        directory: ${{ inputs.build-directory }}
+        verbosity: ${{ inputs.build-verbosity }}
+        enable-code-analysis: ${{ inputs.build-enable-code-analysis }}
+        enable-nuget-caching: ${{ inputs.build-enable-nuget-caching }}
+        enable-npm-caching: ${{ inputs.build-enable-npm-caching }}
+        cache-version: ${{ inputs.build-cache-version }}
+
+    - name: Publish
+      run: |
+        dotnet publish -c Release -o '${{ inputs.build-directory }}/Published'
```
Piedone: Use [`--no-build`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-publish#:~:text=for%20each%20manifest.-,%2D%2Dno%2Dbuild,-Doesn%27t%20build%20the). Also, for CLI parameters always use their full names so the command is better readable (like `--configuration` instead of `-c`).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+        default: 360
+        description: Configuration for the timeout-minutes parameter of the workflow. The 360 is GitHub's default.
+      app-name:
+        required: true
+        type: string
+        description: The name of the existing app service.
+      slot-name:
+        required: true
+        type: string
+        description: Name of the web app slot.
```
Piedone: Add these notes to the descriptions: https://github.com/Azure/webapps-deploy/issues/56#issuecomment-1220019916

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+        enable-nuget-caching: ${{ inputs.build-enable-nuget-caching }}
+        enable-npm-caching: ${{ inputs.build-enable-npm-caching }}
+        cache-version: ${{ inputs.build-cache-version }}
+
+    - name: Publish
+      run: |
+        dotnet publish -c Release -o '${{ inputs.build-directory }}/Published' 
+
+    - name: Login
+      uses: azure/login@24848bc889cfc0a8313c2b3e378ac0d625b9bc16
```
Piedone: See https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=For%20readability%2C%20also%20comment%20the%20version%20there.

Also for the other hashes.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+      with:
+        creds: ${{ secrets.DOTNEST_TENANTS_CORE_EUS_STAGING_AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL }}
```
Piedone: This is a generic workflow, to be used not just by multiple projects of ours but by others too. It shouldn't reference DotNest. Pass in secrets instead (also see my note about `inherit`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+        enable-code-analysis: ${{ inputs.build-enable-code-analysis }}
+        enable-nuget-caching: ${{ inputs.build-enable-nuget-caching }}
+        enable-npm-caching: ${{ inputs.build-enable-npm-caching }}
+        cache-version: ${{ inputs.build-cache-version }}
+
+    - name: Publish
+      run: |
+        dotnet publish -c Release -o '${{ inputs.build-directory }}/Published' 
+
+    - name: Login
```
Piedone:
```suggestion
    - name: Login to Azure
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+      resource-group-name:
+        required: true
+        type: string
+        description: Name of the resource group.
+      url:
+        required: true
+        type: string
+        description: The URL of the web app slot.
+
+jobs:
```
Piedone: Use `pwsh` as the shell (since you didn't specify the shell, under Ubuntu the default bash will be used): https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=Prefer%20to%20use%20PowerShell%207%2B%20scripts. From looking at it, I don't think you'll need to change the scripts.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,149 @@
...
+      with:
+        creds: ${{ secrets.DOTNEST_TENANTS_CORE_EUS_STAGING_AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL }}
+        enable-AzPSSession: true
+
+    - name: Stop Web App Slot
+      uses: Azure/powershell@38b0aa56c413239bd9dcc898d1f62986233658bc
+      with:
+        inlineScript: |
+          Stop-AzWebAppSlot -ResourceGroupName ${{ inputs.resource-group-name }} -Name ${{ inputs.app-name }} -Slot ${{ inputs.slot-name }}
+        azPSVersion: "latest"
```
Piedone: Do you need quotes here, and below for `package`?

Piedone (changes requested):

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -0,0 +1,161 @@
...
+        directory: ${{ inputs.build-directory }}
+        verbosity: ${{ inputs.build-verbosity }}
+        enable-code-analysis: ${{ inputs.build-enable-code-analysis }}
+        enable-nuget-caching: ${{ inputs.build-enable-nuget-caching }}
+        enable-npm-caching: ${{ inputs.build-enable-npm-caching }}
+        cache-version: ${{ inputs.build-cache-version }}
+
+    - name: Publish
+      run: |
+        dotnet publish --configuration Release --output '${{ inputs.build-directory }}/Published' --no-build
```
Piedone: Use `build-verbosity` here too.

Piedone: Don't you just have to publish the web project, not the full solution?

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -103,47 +97,61 @@ jobs:
...
       uses: Azure/powershell@38b0aa56c413239bd9dcc898d1f62986233658bc
       with:
         inlineScript: |
           Stop-AzWebAppSlot -ResourceGroupName ${{ inputs.resource-group-name }} -Name ${{ inputs.app-name }} -Slot ${{ inputs.slot-name }}
-        azPSVersion: "latest" 
+        azPSVersion: latest
 
-    - name: Deploy
+    - name: Deploy to Azure App Service
+      # v2
```
Piedone: This is v2.2.3, not v2. Similarly, be precise with the other versions.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -64,11 +40,11 @@ on:
       app-name:
         required: true
         type: string
-        description: The name of the existing app service.
+        description: What you see at the top of the blade on the Azure Portal. Can contain uppercase letters too.
       slot-name:
         required: true
         type: string
-        description: Name of the web app slot.
+        description: Also what you see at the top of the blade when you open the slot, before the app name in parenthesis.
```
Piedone: This might not be something that people see after reading the description of `app-name`.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -77,11 +53,25 @@ on:
...
+        required: false
+        type: string
+        default: "*.Web.csproj"
+        description: >
+          The path of the web project file to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet
+          as-is, so grepping is still possible but a name with spaces must be escaped separately.
+      runtime:
+        required: false
+        type: string
+        description: If specified, the web app gets published using this runtime and in "ready to run" mode.
```
Piedone: The runtime config is not directly tied to R2R. In TC we actually publish with R2R off but with `self-contained` on, which requires the runtime to be specified (we don't specify `self-contained` but it's implied by setting the runtime).

So, please have configuration to be able to set these three independently, and publish DotNest self-contained.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -90,6 +80,10 @@ jobs:
         timeout-minutes:
         - ${{ inputs.timeout-minutes }}
     timeout-minutes: ${{ matrix.timeout-minutes }}
+    env:
+      publish-command-base: >-
+        dotnet publish (Get-ChildItem ${{ inputs.web-project-path }}).FullName --configuration Release
+        --output '${{ inputs.build-directory }}/Published' --verbosity ${{ inputs.build-verbosity }}
```
Piedone: This is an interesting approach, but my issue is that we're reinventing a lot of wheels from `build-dotnet`. Instead, I'd keep using it and publish with `no-build` (the two together should take basically the same amount of time then just publish with build). Even if needs adding an option to `build-dotnet` to build a csproj instead of a solution.

Piedone (changes requested):

.github/workflows/deploy-to-azure-app-service.yml
DemeSzabolcs: I didn't need to change the logic of `build-dotnet` it already works with `csproj`s. Though I adjusted the descriptions, variable names, and comments to reflect that.

.github/actions/build-dotnet/action.yml
```diff
@@ -41,12 +41,12 @@ inputs:
   dotnet-build-switches:
     required: false
     default: ""
-    description: Additional command line switches given to `dotnet build`. You can split these into multiple lines.
```
DemeSzabolcs, line 44: https://stackoverflow.com/a/59077879/16192320

.github/actions/build-dotnet/action.yml
```diff
@@ -41,12 +41,12 @@ inputs:
...
-  solution-path:
+    description: Additional command line switches given to `dotnet build`. You must split these into multiple lines, even the parameter value.
```
DemeSzabolcs: https://stackoverflow.com/a/59077879/16192320

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -32,6 +32,35 @@ on:
         type: string
         default: quiet
         description: Verbosity parameter for dotnet build.
+      build-enable-code-analysis:
+        required: false
+        type: string
+        default: "true"
```
Piedone: Let's do false for deploys.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -56,14 +85,26 @@ on:
...
-        default: "*.Web.csproj"
+        default: "./src/*Web/*.Web.csproj"
         description: >
           The path of the web project file to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet
           as-is, so grepping is still possible but a name with spaces must be escaped separately.
       runtime:
         required: false
         type: string
-        description: If specified, the web app gets published using this runtime and in "ready to run" mode.
+        description: Publishes the application for a given runtime. If you use this option, use --self-contained or --no-self-contained also.
```
Piedone: Instead of the switches one should use the below parameter though.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -32,6 +32,35 @@ on:
         type: string
         default: quiet
         description: Verbosity parameter for dotnet build.
+      build-enable-code-analysis:
```
Piedone, line 35: Keep to the 120 character limit with descriptions.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -97,47 +134,60 @@ jobs:
...
+    - name: Set up runtime option
       run: |
-        if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj)
-        {
-          Write-Output "::group::Gulp Extensions found. It needs to be explicitly built before the solution."
-
-          $startTime = [DateTime]::Now
-          dotnet build src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj --verbosity ${{ inputs.build-verbosity }}
-          $endTime = [DateTime]::Now
+        $runtime_option = "runtime-option="
```
Piedone: Don't use snake case for variables.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -97,47 +134,60 @@ jobs:
...
+        directory: ${{ inputs.build-directory }}
+        verbosity: ${{ inputs.build-verbosity }}
+        enable-code-analysis: ${{ inputs.build-enable-code-analysis }}
+        enable-nuget-caching: ${{ inputs.build-enable-nuget-caching }}
+        enable-npm-caching: ${{ inputs.build-enable-npm-caching }}
+        cache-version: ${{ inputs.build-cache-version }}
+        solution-or-project-path: ${{ inputs.web-project-path }}
+        dotnet-build-switches: |
+          --self-contained
+          ${{ inputs.self-contained }}
```
Piedone, line 163: Can't be one line?

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -97,47 +134,60 @@ jobs:
...
+        if(![string]::IsNullOrEmpty("${{ inputs.runtime }}"))
+        {
+          $runtime_option = $runtime_option + "--runtime=" + "${{ inputs.runtime }}"
         }
 
-    - name: Build and Publish
-      if: ${{ inputs.runtime == '' }}
-      run: |
-        Invoke-Expression "${{ env.publish-command-base }}"
+        $runtime_option >> $env:GITHUB_ENV
```
Piedone: Use step output instead of an env var to pass this around.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -97,47 +134,60 @@ jobs:
...
+          ${{ inputs.self-contained }}
+          -p:PublishReadyToRun=${{ inputs.ready-to-run }}
+          ${{ env.runtime-option }}
 
+    - name: Publish
+      run: >
+        dotnet publish (Get-ChildItem ${{ inputs.web-project-path }}).FullName --configuration Release --no-build
+        --output '${{ inputs.build-directory }}/Published' --verbosity ${{ inputs.build-verbosity }}
+        --self-contained ${{ inputs.self-contained }} -p:PublishReadyToRun=${{ inputs.ready-to-run }}
+        ${{ env.runtime-option }}
```
Piedone: Indent these and have one line per option for clarity.

Docs/Workflows.md
```diff
@@ -37,10 +37,10 @@ jobs:
 
 ## Build .NET solution workflow
 
-Builds a .NET solution with static code analysis. You can use it along the lines of the following:
+Builds a .NET solution (or project) with static code analysis. You can use it along the lines of the following:
 
 ```yaml
-name: Build and Test
+name: Build
```
Piedone, line 43: Good catches!

.github/actions/build-dotnet/Build-DotNetSolution.ps1
```diff
@@ -48,18 +48,18 @@ if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj
     Write-Output "::endgroup::"
 }
 
-# This prepares the solution with the Lombiq.Analyzers files. The output and exit code are discarded because they will
+# This prepares the solution or project with the Lombiq.Analyzers files. The output and exit code are discarded because they will
```
Piedone: Keep to 120 characters. Also, rename the script too.

.github/actions/build-dotnet/action.yml
```diff
@@ -41,12 +41,12 @@ inputs:
...
+  solution-or-project-path:
     required: false
     default: "*.sln"
     description: >
-      The path of the solution file to be built. If you have exactly one .sln file in the current directory then this
+      The path of the solution or project file to be built. If you have exactly one .sln file in the current directory then this
```
Piedone: Add an example of a project file match here.

Piedone (changes requested):

.github/workflows/deploy-to-azure-app-service.yml
DemeSzabolcs: It can't: https://stackoverflow.com/a/59077879/16192320, this is why I modified the description of this input too.

.github/workflows/deploy-to-azure-app-service.yml
DemeSzabolcs: I can't indent them, since it will not recognize the switches: 
image

Piedone: You need to write PS here, so use backticks to divide single-line commands into multiple lines.

.github/workflows/deploy-to-azure-app-service.yml
Piedone: Add it as a comment.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -0,0 +1,209 @@
...
+        description: >
+          The path of the web project file to be processed. The value is given to PowerShell's `Get-ChildItem` cmdlet
+          as-is, so grepping is still possible but a name with spaces must be escaped separately.
+      runtime:
+        required: false
+        type: string
+        description: Publishes the application for a given runtime.
+      self-contained:
+        required: false
+        type: boolean
```
Piedone: While this would be nicer, use strings, just as for the other bool parameters for consistency. This is necessary because composite actions can only use string parameters. Later we might want to move the deployment to an action too, so this is also useful for future-proofing.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -92,16 +93,17 @@ on:
       runtime:
         required: false
         type: string
-        description: Publishes the application for a given runtime. If you use this option, use --self-contained or --no-self-contained also.
+        description: Publishes the application for a given runtime.
```
Piedone: Keep the note about self-contained, that's important.

Piedone (changes requested):

Piedone (approved):

PR #49, opened 2022-09-07
OSOE-176: Add package validation to NuGet publishing

### LINKED ISSUE #45: Add package validation to NuGet publishing (OSOE-176)
The .NET SDK has [package validation](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview), which are useful checks when you publish your NuGet package. Let's utilize them, and configure them how we see them suitable.

- [Baseline version validator](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/baseline-version-validator):
  - We already have the previous version lookup code in the `publish-nuget` action to get the baseline version. This needs to compare against the latest release (i.e. not pre-release/alpha) version. Note that there might not be a previous release yet, in which case this validation shouldn't run.
  - What to do during alpha releases, releases for issues? Breaking changes should still be tackled, but we need to be able to accept them. The _CompatibilitySuppressions.xml_ file is good for this, but this also needs to be automatically removed upon a major release (i.e. when we may introduce breaking changes).
- [Compatible runtime validator](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/compatible-framework-validator) looks good as it is.
- [Compatible framework validator](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/compatible-framework-in-package-validator) looks good as it is.
- Do everything from the CLI if possible, without the need to modify csprojs.

Piedone:
OSOE-176

Fixes #45.

[Example when it runs](https://github.com/Lombiq/UI-Testing-Toolbox/runs/8247531935?check_suite_focus=true)

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -159,24 +159,24 @@ runs:
             "-p:EnablePackageValidation=true"
         )
 
-    #- name: Push with dotnet
-    #  shell: pwsh
-    #  run: dotnet nuget push artifacts/*.nupkg --api-key $Env:API_KEY --source '${{ steps.setup.outputs.source-url }}' --skip-duplicate
+    - name: Push with dotnet
+      shell: pwsh
+      run: dotnet nuget push artifacts/*.nupkg --api-key $Env:API_KEY --source '${{ steps.setup.outputs.source-url }}' --skip-duplicate
 
-    #- name: Publish Artifacts
-    #  uses: actions/upload-artifact@v3
-    #  with:
-    #    name: NuGet-Package
-    #    path: artifacts
+    - name: Publish Artifacts
+      uses: actions/upload-artifact@v3
+      with:
+        name: NuGet-Package
+        path: artifacts
 
-    ## This is not in its own action because it's all very specific to this NuGet publishing.
-    #- name: Create Release
-    #  # v1.10.0
-    #  uses: ncipollo/release-action@58ae73b360456532aafd58ee170c045abbeaee37
-    #  # This is to prevent creating releases when pushing tags for issue-specific pre-releases like
-    #  # v4.3.1-alpha.osoe-86.
-    #  if: "!contains(steps.setup.outputs.publish-version, '-')"
-    #  with:
-    #    allowUpdates: true
-    #    generateReleaseNotes: true
-    #    artifacts: artifacts/*.nupkg, artifacts/*.snupkg
+    # This is not in its own action because it's all very specific to this NuGet publishing.
+    - name: Create Release
+      # v1.10.0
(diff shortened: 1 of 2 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -124,25 +124,26 @@ runs:
...
+      # * p:EnablePackageValidation=true is for package validation, see:
+      #   https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview.
       run: |
         if ([string]::IsNullOrWhiteSpace($Env:API_KEY)) {
           Write-Output "::error::API_KEY is missing or empty."
           exit 1
         }
 
-        $noWarn = @('NU5104') + @'
+        $noWarn = @('NU5104%3BCS1573%3BCS1591%3BVSTHRD002%3BVSTHRD200') + @'
```
dministro, line 139: Why these new suppressions are required here?
I didn't find any information about this under the issue.

.github/actions/publish-nuget/action.yml
```diff
@@ -154,7 +155,8 @@ runs:
             "-p:WarnOnPackingNonPackableProject=True",
             "-p:IncludeSymbols=${{ inputs.dotnet-pack-include-symbols }}",
             "-p:SymbolPackageFormat=snupkg",
-            "-p:NoDefaultExcludes=true"
+            "-p:NoDefaultExcludes=true",
+            "-p:EnablePackageValidation=true"
```
dministro, line 159: No baseline version validation here(as you mentioned under the issue) but under the related [wiki page](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/786857987/Managing+releases+of+open-source+projects#:~:text=To%20check%20for%20breaking%20changes%20in%20the%20.) the `-p:PackageValidationBaselineVersion=$current` is present.
Why the difference?

.github/actions/publish-nuget/action.yml
Piedone: These are the same as the build ones, because I removed `--no-build` from the `pack` command. That was necessary, since, apparently, `-p:EnablePackageValidation=true` invokes the "Build" target, see https://github.com/Lombiq/UI-Testing-Toolbox/runs/8234626733?check_suite_focus=true#step:19:140

.github/actions/publish-nuget/action.yml
Piedone: See https://github.com/Lombiq/GitHub-Actions/issues/45#issuecomment-1240568296

dministro: I saw your comment under #45  and I understand why `p:PackageValidationBaselineVersion` isn't here, my question is why it is required under the wiki.
But now it's clear. (I read the wiki again)

dministro (approved):

PR #50, opened 2022-09-15
OSOE-173: Automatically create Jira issues for GitHub community activity

Piedone:
OSOE-173


Docs/Workflows.md
```diff
@@ -99,3 +99,13 @@ jobs:
...
 Refer to [Github Actions reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#overview) for more information.
+
+# Jira issue-creating for community activities workflow
+
+Creates Jira issues for community activities happening on GitHub, like issues, discussions, pull requests being opened.
+
+Set up secrets for the parameters as explained [here](https://github.com/marketplace/actions/jira-login#enviroment-variables). You may use secret names without the `DEFAULT_` prefix, but that's our recommendation so you can override these on a per-repository basis.
+
+```yaml
+placeholder
```
Piedone: Will add once the workflow is finalized.

.github/actions/publish-nuget/action.yml
```diff
@@ -27,9 +27,6 @@ inputs:
       The desired NuGet package version used for publishing. If not specified, the GITHUB_REF_NAME environment variable
       is used which is suitable if the version is derived from a git tag. If "USE_GITHUB_RUN_NUMBER" is used, then the
       version is derived from the latest non-prerelease version tag and the current run number.
-secrets:
-  API_KEY:
-    required: true
```
Piedone, line 32: This didn't actually do anything, and it's also invalid in the schema.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
```diff
@@ -0,0 +1,50 @@
+param ($GitHub)
+
+switch ($GitHub.event_name)
+{
+    "discussion"
+    {
+        $summary = "Respond to `"$($GitHub.event.discussion.title)`""
+        $description = $Env:DISCUSSION_JIRA_ISSUE_DESCRIPTION
```
sarahelsaig: Why is this an environment variable instead of a parameter? Seems inconsistent.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
```diff
@@ -0,0 +1,50 @@
...
+        $link = $GitHub.event.discussion.html_url
+    }
+    "issues"
+    {
+        $summary = "$($GitHub.event.issue.title) in $($GitHub.repository)"
```
sarahelsaig: Shouldn't this be quoted too? Alternatively the `in $($GitHub.repository)` could be parenthesized.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
```diff
@@ -0,0 +1,50 @@
...
+            {
+                $type = "Bug"
+            }
+            elseif ($labelName -eq "enhancement")
+            {
+                $type = "New Feature"
+            }
+
+            $i++
+        }
```
sarahelsaig: Looks like this is actually a `foreach` that breaks if `$null -ne $type`.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
```diff
@@ -0,0 +1,50 @@
+param ($GitHub)
+
+switch ($GitHub.event_name)
```
sarahelsaig, line 15: Don't you need a default case for this? Or to check if `$summary` is null or empty at the end before creating outputs.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -0,0 +1,103 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Scripts
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+        cd ${{ github.action_path }}
+        cd ../
+        "$((Get-Location).Path)/verify-submodule-pull-request" >> $Env:GITHUB_PATH
```
sarahelsaig: Why not join the path instead?
`Join-Path -Resolve "${{ github.action_path }}" .. verify-submodule-pull-request`

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -0,0 +1,103 @@
...
+
+    # Don't do anything for pull requests that are already related to an issue, i.e. their titles start with an issue
+    # key.
+    - name: Check for Jira Issue-Related Pull Request
+      id: issue-related-pr
+      shell: pwsh
+      run: |
+        Write-Output "::set-output name=is-not-issue-pr::$(-not(Check-PullRequestTitle ${{ github.event.pull_request.title }}))"
```
sarahelsaig: This accidentally works, because we only need the first word anyway. But generally the `${{ ... }}` is preprocessor substitution, so you ought to wrap it in quotes.

sarahelsaig: I think the negations make things confusing and using `steps.issue-related-pr.outputs.is-issue-pr == 'False'` would be cleaner.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -0,0 +1,103 @@
...
+        JIRA_BASE_URL: ${{ env.JIRA_BASE_URL }}
+        JIRA_USER_EMAIL: ${{ env.JIRA_USER_EMAIL }}
+        JIRA_API_TOKEN: ${{ env.JIRA_API_TOKEN }}
+
+    - name: Initialize Jira Issue Details
+      if: steps.issue-related-pr.outputs.is-not-issue-pr == 'True'
+      id: issue-details
+      shell: pwsh
+      run: |
+        # The funny indentation is necessary for here-strings.
```
sarahelsaig: Funny?

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -0,0 +1,103 @@
...
+      with:
+        project: ${{ env.JIRA_PROJECT_KEY }}
+        issuetype: ${{ steps.issue-details.outputs.type }}
+        summary: ${{ steps.issue-details.outputs.summary }}
+        description: ${{ fromJSON(steps.issue-details.outputs.jsonDescription) }}
+        fields: ${{ steps.issue-details.outputs.fields }}
+
+    - name: Add Remote Link to Jira Issue
+      if: steps.issue-related-pr.outputs.is-not-issue-pr == 'True'
+      shell: bash
```
sarahelsaig: How come this is bash?

Docs/Workflows.md
```diff
@@ -99,3 +99,24 @@ jobs:
 If this is for a submodule of [Lombiq's Open-Source Orchard Core Extensions](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/), the `repo` input can be omitted, because the above is its default value. Otherwise, use your parent repository's address in the `{owner}/{repo_name}` format.
 
 Refer to [Github Actions reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#overview) for more information.
+
+## Jira issue-creating for community activities workflow
```
sarahelsaig:
```suggestion
## Jira issue creation for community activities workflow
```

Docs/Workflows.md
```diff
@@ -99,3 +99,24 @@ jobs:
...
+
+Creates Jira issues for community activities happening on GitHub, like issues, discussions, pull requests being opened.
```
sarahelsaig: We should explain what makes the activities "community". Seems to me the distinction only exists for pull requests anyway. In other words this creates Jira issues for any issue or discussion and for pull requests that don't already contain a Jira issue code.

Docs/Workflows.md
```diff
@@ -99,3 +99,24 @@ jobs:
...
+
+Set up secrets for the `JIRA_` parameters as explained [here](https://github.com/marketplace/actions/jira-login#enviroment-variables). You may use secret names without the `DEFAULT_` prefix, but that's our recommendation for organization-level secrets, so you have defaults but can override them on a per-repository basis.
+
+The secrets with the `_JIRA_ISSUE_DESCRIPTION` suffix should contain templates for the Jira issues to be created, using the internal markup format of Jira (not Markdown). Example for one for `ISSUE_JIRA_ISSUE_DESCRIPTION`:
```
sarahelsaig: Shouldn't this be `JIRA_*`, `DEFAULT_*` and `*_JIRA_ISSUE_DESCRIPTION`? Especially the latter looks weird to me as it currently is.

.github/workflows/create-jira-issues-for-community-activities.yml
```diff
@@ -0,0 +1,43 @@
...
+      JIRA_API_TOKEN:
+        required: true
+      JIRA_PROJECT_KEY:
+        required: true
+      DISCUSSION_JIRA_ISSUE_DESCRIPTION:
+        required: true
+      ISSUE_JIRA_ISSUE_DESCRIPTION:
+        required: true
+      PULL_REQUEST_JIRA_ISSUE_DESCRIPTION:
+        required: true
```
sarahelsaig: Can't we provide sensible defaults for these? I understand that you don't want to include our issue completion checklist URL, but is that applicable for the DISCUSSION and PULL_REQUEST types too? Also we could provide a default without the offending link for issue description anyway, in case anyone else wants to use the workflow.

sarahelsaig (changes requested):

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
Piedone: This is adding the repo to the title. Perhaps it should rather add the configured component, or no suffix at all. What do you think?

sarahelsaig: Well, the issue component is optional. So it could be the component if available and the repo otherwise.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
Piedone: I wanted to minimize passing around secrets.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
Piedone: Good idea. I'll actually add this to the other titles too.

Piedone: Do you mean quotes around `$($GitHub.repository)`? Since it's a clearly self-contained text (just as the component name), unlike a PR or discussion title, it didn't feel right.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
Piedone: I knew you'll notice this but I was like "eh, who calls this for some other event deserves their fate".

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: image

Piedone: Yeah, initially the output was meant to be used with any equality check, but then later it turned out that `if` won't understand them as Booleans anyway.

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: It perhaps depends on your humor :D. Meaning that the second and third lines here are not indented:

```powershell
        $github = @"
        ${{ toJSON(github) }}
        "@ | ConvertFrom-Json
```

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: Curl was the easiest way to do this, and all the Jira API examples were using curl from bash. But I changed it to PS.

Docs/Workflows.md
Piedone: It's just that activities on a public GH project for OSS software = community activity. For us two there's indeed the connotation that it's only "community" activity if not done by a Lombiq team member. I modified the description so it's sufficiently clear for anybody outside of Lombiq too.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
sarahelsaig: You could still pass those environment variables as parameters in the action where this script is called. Just to avoid depending on hidden variables.

.github/actions/create-jira-issues-for-community-activities/action.yml
sarahelsaig: It's still not entirely clear. All three lines are indented to the first column of this yaml block. The additional indentation is uniformly chopped by the yaml deserializer before the shell interpreter gets the script so as far as the here-string is concerned you just have
```pwsh
$github = @"
${{ toJSON(github) }}
"@ | ConvertFrom-Json
```
If the "funny" is that the closing `"@` of the here-string must be at the line start, then ok. By the way you are allowed to further indent the `${{ toJSON(github) }}` line if you think it will improve readability. That whitespace is ignored during deserialization anyway.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -75,23 +80,23 @@ runs:
...
+          --user '${{ env.JIRA_USER_EMAIL }}:${{ env.JIRA_API_TOKEN }}' `
+          --header 'Accept: application/json' `
+          --header 'Content-Type: application/json' `
           --data '{
-            "object": {
-                "url":"${{ steps.issue-details.outputs.link }}",
-                "title":"GitHub resource"
-            }
-        }'
+                \"object\": {
```
sarahelsaig, line 106: Why is this backslashed now?

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
```diff
@@ -1,42 +1,50 @@
...
-        $summary = "Review `"$($GitHub.event.pull_request.title)`""
+        $summary = "Review `"$($GitHub.event.pull_request.title)`"$titleSuffix"
         $description = $Env:PULL_REQUEST_JIRA_ISSUE_DESCRIPTION
         $link = $GitHub.event.pull_request.html_url
     }
+    default
+    {
+        $message = "Unknown event `"$($GitHub.event_name)`". Please only call this script for one of the following " +
+            "events: discussion, issues, pull_request."
+        Write-Error $message
```
sarahelsaig: Let's make it an [error annotation](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message).
```suggestion
        Write-Error "::error::$message"
        exit 1
```

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: It's necessary in PS.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
sarahelsaig: I mean the other summaries are like ``"Respond to `"$($GitHub.event.discussion.title)`""``, so I wondered if this should be ``$summary = "`"$($GitHub.event.issue.title)`" in $($GitHub.repository)"`` as well.

The alternative I mentioned would be adding a parenthesis around the repo.
So in the new code `$titleSuffix = " (in $context)"`.

Docs/Workflows.md
```diff
@@ -99,3 +99,26 @@ jobs:
...
+```text
+h1. Summary
+See the linked GitHub issue, including all the comments.
+
+h1. Checklist
+* Assign yourself to the referenced GitHub issue.
+* [Issue completion checklist|https://example.com/checklist]
+```
+
+All three templates are optional and if not provided, defaults will be used.
```
sarahelsaig, line 139: [The workflow inputs are still required though.](https://github.com/Lombiq/GitHub-Actions/pull/50/files#diff-724e9168a79d65edb0fe14c60f55080e9e2f8a5090e654273248f4e9e9156ebbR15-R20)

sarahelsaig (changes requested):

.github/actions/create-jira-issues-for-community-activities/action.yml
sarahelsaig: Well, as long as it works...

Docs/Workflows.md
Piedone: I'm not done yet ;).

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: Let's just remove this comment, it's really not that crucial.

.github/workflows/create-jira-issues-for-community-activities.yml
Piedone: This was a fun hour wasted, but apparently there's no way to override env vars GitHub sets. Anyway, I've did it otherwise.

.github/actions/create-jira-issues-for-community-activities/action.yml
Piedone: It does, and it certainly didn't work without it :D.

Docs/Workflows.md
sarahelsaig: Sorry.

.github/actions/create-jira-issues-for-community-activities/Initialize-IssueDetails.ps1
Piedone: I see. It's not `$summary = "`"$($GitHub.event.issue.title)`" in $($GitHub.repository)"` because, unlike the other two events, we're opening a Jira issue for a GitHub **issue**, so it's adequate to use the same title. Note that while in the case of issues the GH title is part of the title proper, in the other two cases the PR/discussion titles are references _from_ the title:

image

The parenthesis, I'm not sure. It doesn't seem to be necessary for me.

sarahelsaig (changes requested):
I've just noticed, that it creates a completely new Jira issue when you edit the git issue's title or description. ([see here](https://github.com/Lombiq/Community-Activities-Notification-Test/issues/38)) I think the action should not run if the title already has a Jira issue code in it.

Piedone:
This is deliberate temporarily. The [workflow also triggers on issue edit and labeled](https://github.com/Lombiq/Community-Activities-Notification-Test/blob/dev/.github/workflows/create-jira-issues-for-community-activities.yml#L7) so Jira issues will be automatically created for existing GH issues when we edit them or add a label. This is just for:

> Edit all issues in the affected projects that don’t have a corresponding Jira issue yet (just adding a space in the text or something invisible like that, or a label if necessary and if it triggers too) to run the workflow. Once done, replace types: [opened, edited, labeled] with types: opened in all workflows. Note that closing an issue doesn’t trigger an edit, so stale issues don’t need to be closed first. 

If the GH project in question didn't have any issues, then I only added a trigger for the `opened` event ([like here](https://github.com/Lombiq/NPM-Targets/pull/24/files#diff-724e9168a79d65edb0fe14c60f55080e9e2f8a5090e654273248f4e9e9156ebbR7)), which won't trigger when you edit anything of an existing issue.

sarahelsaig:
imo that's a bad solution, because editing the issue description is not rare at all. I do it often in OCC so I wouldn't use this there even as a temporary measure. Like I said, checking for issue code in the title (or rather, if the title ends with `(PROJ-000)`) would do the same without causing crippling side effects. Right now if a community contributor edits a GH issue, a new JIRA issue is created. If they get confused by the change or edit the issue for any other reason then another JIRA issue is spawned. Seems like a vulnerability to me.

Piedone:
This is for those 3 minutes until I go through all the unprocessed issues, do some noop edit, wait for the new Jira issues to be created, and change the workflow to just trigger on `opened`.

Piedone:
I can also do the merges, to minimize the time window and thus chance of anybody editing anything between the merge and me changing the triggers.

PR #56, opened 2022-09-20
OSOE-156: Optimize the file management of Windows builds

### LINKED ISSUE #32: Optimize the file management of GitHub Actions Windows builds (OSOE-156)
The Windows builds of our GitHub Actions workflows are always a lot slower than the Ubuntu ones for every operation, even checkout (at least 2-3x slower, all the way up to 40x sometimes). Let’s try to make this better by tackling file management-related Windows issues.

- See [this thread](https://github.community/t/github-actions-windows-runner-an-order-of-magnitude-slower-than-the-ubuntu-one-why/261186/3) for a discussion about this with various tips. But this thread is now lost due to GitHub deleting the community forum, see: https://github.com/orgs/community/discussions/30387
- Check if we can configure Windows Defender and if it makes GHA builds faster: Check if under Windows we can configure Windows Defender to exclude devenv.exe, MSBuild.exe, and dotnet.exe from the checked processes, as well as the folder of the checkout directory, and if that makes builds faster.
- Related issues:
  - https://github.com/actions/runner-images/issues/5166
  - https://github.com/actions/runner-images/issues/3577#issuecomment-990979789
  - https://github.com/actions/setup-dotnet/issues/260
  - https://github.com/NuGet/Home/issues/11548
- [Windows with a better filesystem](https://superuser.com/questions/240731/windows-with-a-better-filesystem)

### To be checked

See the attached spreadsheet for measurements.

- [x] Disabling Windows Defender by @0liver
  - This doesn't seem to have any noticeable effect, and especially not a positive one.
- [x] "While there is no ext4 driver for Windows there is [WinBtrfs](https://github.com/maharmstone/btrfs) for [Btrfs](https://en.wikipedia.org/wiki/Btrfs), another major Linux file system. Supposedly it’s better than NTFS though worse than ext4. None of the other major Linux file systems (xfs, f2fs, reiserfs, reiser4) have Windows drivers." by @DAud-IcI
  - Tried WinBtrfs out on [this commit](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/205/commits/afea9090fe80343c3609dc5f070018dd3d9bcee7), see measurements in the spreadsheet. If anything, it's slower by about 10%. [Corresponding GHA commit](https://github.com/Lombiq/GitHub-Actions/pull/56/commits/6322eef246b78cc1f58875e78f9abb9d8b8a24f6).
- [x] NTFS settings:
  - [x] "On NTFS playing around with fsutil [might be helpful](https://gist.github.com/xnohat/0359165197b5f690b438c96861e8041a), though I assume the VM maintainers have already did this. If not, should open a PR on [actions/virtual-environments](https://github.com/actions/virtual-environments/issues/) and request it." by @DAud-IcI.
     - Needs reboot, so should rather be done in the image. Since **rebooting a GitHub-hosted runner doesn’t work** (see [this run](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3092925275/jobs/5004730033)), none of these settings can be changed. There doesn't seem to be a way to run startup scripts or anything like that.
     - Shortnames are explicitly [enabled in the GHA image](https://github.com/actions/runner-images/blob/8481b2c846bd93b5665352e665107f08cc209722/images/win/scripts/ImageHelpers/VisualStudioHelpers.ps1#L31).
     - See [this script](https://github.com/Lombiq/GitHub-Actions/blob/a7b707d366a232fb7b790bc7dca966288584f2aa/.github/actions/tune-windows-performance/Initialize-WindowsSettings.ps1) that I used to set NTFS settings.
  - [x] [Similar config recommendations](https://superuser.com/a/488305/47254).
- [x] "As for git performance, looking for a different command line client based on libgit2 might help. (actions/checkout@v3 just front-ends normal git) Or perhaps the [GitHub CLI](https://github.com/cli/cli) which uses a different re-implementation called go-git." by @DAud-IcI 
  - The checkout is slower than under Ubuntu (though lately it seems to be "just" at most 50% slower, and usually not too much slower) it's still just 30s-1m out of a 16-30m build. Even if we optimize this a lot, it won't really matter. So, I don't think we should put any effort into reinventing the git wheel, which seems like quite a lot of work anyway.
- [x] Virtual file system with better "lot of small files" support
  - [Solid File System](https://solid-file-system.software.informer.com/) was perhaps something like this, though it doesn't seem to be an active project. It seems the successor is [CBFS Storage](https://www.callback.com/cbfsstorage/). It [can create virtual drives too](https://www.callback.com/kb/articles/product-comparison.rst). It's not open-source though, so we can't include it in our open-source actions (a license is $1500).
  - [WSL 2 can mount ext4 drives](https://learn.microsoft.com/en-us/windows/wsl/wsl2-mount-disk), including a VHDX file.
    - This is available under Windows 11 Insider Build but [can also be installed from the Microsoft Store too](https://devblogs.microsoft.com/commandline/a-preview-of-wsl-in-the-microsoft-store-is-now-available/). You still need Windows 11, though (the indicated minimum requirement of "Windows 10 version 22000 or higher" is actually Windows 11).
    - [Here's a blog post](https://joeferguson.me/adding-another-disk-to-wsl2/) about how to create a VHDX file to begin with and how to mount it. [Here's an answer](https://github.com/microsoft/WSL/discussions/5896#discussioncomment-1010426) about the same thing with CLI samples.
    - [`setup-wsl` action](https://github.com/Vampire/setup-wsl). Note that this is only WSL 1 because [the currently used Azure VMs don't support nested virtualization](https://github.com/Vampire/setup-wsl/issues/28).
    - [Comment about manual setup](https://github.com/actions/runner-images/issues/50#issuecomment-663844380) but the WSL enable shouldn't be necessary anymore. [PS-based installation is here.](https://github.com/microsoft/WSL/issues/5014#issuecomment-692432322)
    - [Server installation](https://learn.microsoft.com/en-us/windows/wsl/install-on-server) for WSL.
    - See [this script](https://github.com/Lombiq/GitHub-Actions/blob/8c40ef7f4c16a768fa6407c35019e0f148f0502b/.github/actions/tune-windows-performance/Initialize-Ext4Drive.ps1) with basics of WSL drive mounting. However, this can be nicer by using `New-VHD` and `Mount-VHD`, which need Hyper-V enabled (which is necessary for WSL 2 anyway).
    - **WSL 2 is not possible to use under GitHub-hosted runners.**
- [x] Other performance increase options?
  - Nobody seems to care much about .NET build perf on GHA.
  - We've tried NuGet and NPM caching before without benefits.
  - [BuildJet](https://github.com/marketplace/buildjet) seems interesting, but the thing they don't offer is Windows builds.
  - No other ideas...

[Measurements.xlsx](https://github.com/Lombiq/GitHub-Actions/files/9619662/Measurements.xlsx)

### Conclusion

Summary of findings:
- Disabling Windows Defender doesn't help.
- Using a pluggable virtual file system (WinBtrfs) is slower.
- Due to not being able to reboot, we can't change NTFS settings.
- Git performance is not a big issue.
- Perhaps ext4 would be better as a virtual file system, but it's impossible to use on GitHub-hosted runners.

Next steps:
- Tuning NTFS settings can work with self-hosted runners, using our own image. Mounting an ext4 virtual drive with WLS 2 too, for that additionally we need to use an Azure VM type that allows virtualization. I've written scripts for these (or the basics of these) that may be used during packing our own images.
- BuildJet sounds promising, and I'm testing it, but it's only for Ubuntu.

However, the best option, really, is to just use Ubuntu if possible. (It is _possible_ for all of our Orchard Core projects, but for OSOCE, specifically, we need Windows builds to verify cross-platform behavior).

Piedone:
OSOE-156

Fixes #32.


Piedone:
Nothing worked here, unfortunately.

PR #62, opened 2022-10-04
OSOE-156: Code styling

Piedone:
OSOE-156

### DIFF
--- a/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
+++ b/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
@@ -11,5 +11,5 @@ else
         --env 'SA_PASSWORD=Password1!' `
         --publish 1433:1433 `
         --detach 'mcr.microsoft.com/mssql/server:2019-latest' &&
-    docker exec --user 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
+    docker exec --user 0 sql2019 bash -c 'mkdir /data; chmod 777 /data --recursive; chown mssql:root /data'
 }
--- a/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
+++ b/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
@@ -1,10 +1,14 @@
-﻿if ($Env:RUNNER_OS -eq "Windows")
+if ($Env:RUNNER_OS -eq "Windows")
 {
     choco install sql-server-express --no-progress
 }
 else
 {
     docker pull mcr.microsoft.com/mssql/server &&
-    docker run --name sql2019 -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password1!' -p 1433:1433 -d 'mcr.microsoft.com/mssql/server:2019-latest' &&
-    docker exec -u 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
+    docker run --name sql2019 `
+        --env 'ACCEPT_EULA=Y' `
+        --env 'SA_PASSWORD=Password1!' `
+        --publish 1433:1433 `
+        --detach 'mcr.microsoft.com/mssql/server:2019-latest' &&
+    docker exec --user 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
 }
--- a/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
+++ b/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
@@ -4,10 +4,11 @@ if ($Env:RUNNER_OS -eq "Windows")
 }
 else
 {
-    # Commands taken from https://github.com/ankane/setup-sqlserver.
-    wget -qO- https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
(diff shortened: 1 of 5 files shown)


.github/actions/setup-sql-server/Initialize-SqlServer.ps1
```diff
@@ -1,10 +1,14 @@
...
 else
 {
     docker pull mcr.microsoft.com/mssql/server &&
-    docker run --name sql2019 -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=Password1!' -p 1433:1433 -d 'mcr.microsoft.com/mssql/server:2019-latest' &&
-    docker exec -u 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
+    docker run --name sql2019 `
+        --env 'ACCEPT_EULA=Y' `
+        --env 'SA_PASSWORD=Password1!' `
+        --publish 1433:1433 `
+        --detach 'mcr.microsoft.com/mssql/server:2019-latest' &&
```
0liver: If you want to put each argument on a new line, let's start with the first one:
```suggestion
    docker run `
        --name sql2019 `
        --env 'ACCEPT_EULA=Y' `
        --env 'SA_PASSWORD=Password1!' `
        --publish 1433:1433 `
        --detach 'mcr.microsoft.com/mssql/server:2019-latest' &&
```

0liver: But then again I thought that we'd switch to using maps for parameters wherever necessary to avoid the barely visible backticks, like e.g. in [`build-dotnet`](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/build-dotnet/action.yml#L128).

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
```diff
@@ -1,10 +1,14 @@
...
+    docker exec --user 0 sql2019 bash -c 'mkdir /data; chmod 777 /data -R; chown mssql:root /data'
```
0liver: Use the long name for `-c`, too.

0liver (changes requested):

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
Piedone: I really wanted to, but there doesn't seem to be a long name for that. I just noticed that I missed `chmod -R` though.

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
Piedone: I'd argue that if they're barely visible then one needs to increase the font size :D. Jokes aside, in the linked case I can understand, because it's YML mixed with PS mixed with GitHub expressions. This here is a super plain command invocation though.

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
0liver: 😳 Wow - bash must be really old then!

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
0liver: I'll take your word for it. Let's just hope that no-one will waste any time on debugging a missing `` ` `` here.

0liver (approved):

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
sarahelsaig: Not bash, that has only a handful built-in commands. `chmod`, `chown` and others are standalone programs part of the GNU project... which is 39 years old!!

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
sarahelsaig: I think the problem isn't that you don't see the backtick but that you don't see if there is a space or other non-printing character between it and the newline. Also you can never know if some parsers crush the code into a single line before interpreting anyway, then what the backtick escapes is completely implementation-dependent. These can lead to some debugging horror which is why I avoid escaping newlines in every language.

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
0liver: 👀 

I was referring to the lack of a long-name option for `-c` - I would expect `--command` as an equivalent, which is kind of today's standard (at least in the .NET world, I guess).

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
0liver: Ah, thanks, that was the culprit! I couldn't quite remember what it was, but I felt uneasy about it.

I'd vote for switching to argument maps again then. Also, to make a habit out of it and ease copy'n'pasting for colleagues or other random folks on the internet.

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
sarahelsaig: Oh right, sorry. Bash is a couple years newer, but it's dedicated to have a POSIX mode and be fully backwards compatible with Bourne shell (`sh`) which is even older than GNU itself (43 years old). Who needs such a legacy mode today? No idea.

0liver: 🧾 Thanks for the free history lesson! 😎

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
Piedone: I'll reopen this, then.

PR #63, opened 2022-10-06
OSOE-156: Code styling

Piedone:
OSOE-156

### DIFF
--- a/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
+++ b/.github/actions/setup-sql-server/Initialize-SqlServer.ps1
@@ -5,11 +5,15 @@ if ($Env:RUNNER_OS -eq "Windows")
 else
 {
     docker pull mcr.microsoft.com/mssql/server &&
-    docker run `
-        --name sql2019 `
-        --env 'ACCEPT_EULA=Y' `
-        --env 'SA_PASSWORD=Password1!' `
-        --publish 1433:1433 `
-        --detach 'mcr.microsoft.com/mssql/server:2019-latest' &&
+
+    $dockerRunSwitches = @{
+        name = sql2019
+        env = 'ACCEPT_EULA=Y'
+        env = 'SA_PASSWORD=Password1!'
+        publish = 1433:1433
+        detach = 'mcr.microsoft.com/mssql/server:2019-latest'
+    }
+    docker run @dockerRunSwitches &&
+
     docker exec --user 0 sql2019 bash -c 'mkdir /data; chmod 777 /data --recursive; chown mssql:root /data'
 }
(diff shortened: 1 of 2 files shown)


0liver:
What an adventure!

.github/actions/setup-sql-server/Initialize-SqlServer.ps1
```diff
@@ -4,12 +4,17 @@ if ($Env:RUNNER_OS -eq "Windows")
 }
 else
 {
+    # For whatever reason, except name, the switches need to only include a single dash despite them needing a double
+    # dash when executing directly from the command line.
```
0liver: Please remove or rephrase the comment.

0liver (changes requested):

Piedone:
It was a huge waste of time fighting with the peculiarities of how splatting works.

0liver:
You might have been able to test that locally - not the docker command, just the argument splatting.

Anyway, you are smarter now! 🎉

It's probably worth sharing this hard-won knowledge in our wiki on some PS specific page.

Piedone:
I did what I can locally, but most of the stuff was related to Docker (under Ubuntu). I'll add this to the wiki.

sarahelsaig:
You don't have to test it with docker tho.
image

Piedone:
That's what I thought, but hash table-based splatting doesn't work with anything non-PS. Array-based splitting does, but I could get that right in one go, after spending half an hour figuring out why hash table-based splatting won't work the same as with a PS script.

sarahelsaig:
You can see from my screenshot that splatting a hash table prepends the keys with a single dash. So I wouldn't expect it to work with `docker` that expects longform arguments to have the GNU style double dashes.

Piedone:
That's what I did, except for `--name` it does need double dashes too. [This code](https://github.com/Lombiq/GitHub-Actions/blob/8022fce306fa8eb51e2b972e2b77e8f72102f230/.github/actions/setup-sql-server/Initialize-SqlServer.ps1) with [this build](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3197976199/jobs/5222126951) is where I had to give up hash tables.

But please let's not waste any more time on this whole thing.

sarahelsaig:
What you linked actually results in `docker run ---name sql2019 --env 'ACCEPT_EULA=Y SA_PASSWORD=Password1!' --publish '1433:1433' --detach 'mcr.microsoft.com/mssql/server:2019-latest'`. Note the triple dash. Like I said, splatting a table prepends one dash to the key. It's a real misstep from Microsoft because if it prepended double dash then it would be more compatible.
image

It appears that you've given up right before removing the final double dash from your hash table.
image

0liver (approved):

PR #67, opened 2022-10-12
OSOE-396: Only run setup-dotnet if the image doesn't contain the same .NET version

### LINKED ISSUE #66: Only run setup-dotnet if the image doesn't contain the same .NET version (OSOE-396)
Running `actions/setup-dotnet` takes ~10s under Ubuntu, and ~45s under Windows. We always use it to install version `6.0.*`, which is actually pre-installed both [under Ubuntu](https://github.com/actions/runner-images/blob/ubuntu20/20221002.2/images/linux/Ubuntu2004-Readme.md#net-core-sdk) and [Windows](https://github.com/actions/runner-images/blob/win22/20221002.2/images/win/Windows2022-Readme.md#net-core-sdk). Let's not run the installation when the desired version exists.

Also, let's upgrade `actions/setup-dotnet` to v3, perhaps that already handles this.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-396)

Piedone:
OSOE-396
Fixes #66


.github/actions/setup-dotnet/action.yml
```diff
@@ -1,16 +1,16 @@
...
     required: false
-    default: 6.0.*
+    default: 6.0.x
     description: Version of the .NET SDK to set up.
 
 runs:
   using: "composite"
   steps:
-    - name: Set environment variables
+    - name: Set Environment Variables
```
sarahelsaig:
```suggestion
    - name: Set Environment Variables
```
Fix spacing.

PR #69, opened 2022-10-13
OSOE-385: verify-dotnet-consolidation action should not fail the installation if it's already installed

### LINKED ISSUE #58: verify-dotnet-consolidation action should not fail the installation if it's already installed (OSOE-385)
And ideally, it should also replace the installed version with the desired one.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-385)

BenedekFarkas:
OSOE-385
Fixes #58


.github/actions/verify-dotnet-consolidation/action.yml
```diff
@@ -20,7 +20,8 @@ runs:
   steps:
     - name: Setup
       shell: pwsh
-      run: dotnet tool install dotnet-consolidate --global --version 2.0.0
+      run: |
+        & "${{ github.action_path }}/../../../Scripts/Install-DotNetTool.ps1" -Name "dotnet-consolidate" -Version "2.0.0" -Global
```
Piedone: Now that we have a Scripts folder, please also move Check-PullRequestTitle.ps1 to it.

Scripts/Install-DotNetTool.ps1
```diff
@@ -0,0 +1,48 @@
...
+    [string]
+    $Name,
+
+    [Parameter(Mandatory = $true, HelpMessage = "The version of the tool to install.")]
+    [string]
+    $Version,
+
+    [Parameter(HelpMessage = "When present, the tool will be installed globally, locally otherwise.")]
+    [switch]
+    $Global
```
Piedone, line 12: This can default to being on, since we'll always install tools globally when used during a workflow.

Scripts/Install-DotNetTool.ps1
```diff
@@ -0,0 +1,48 @@
...
+    }
+} | Where-Object { $_.PackageId -eq $Name }
+        
+$doInstall = $true
+        
+if ($null -ne $installedTool -and $installedTool -ne "")
+{
+    if ($installedTool.Version -ne $Version)
+    {
+        dotnet tool uninstall $Name $scopeString
```
Piedone, line 36: Won't a `dotnet isntall` override the isntalled version anyway?

Piedone (changes requested):

Scripts/Install-DotNetTool.ps1
```diff
@@ -0,0 +1,48 @@
...
+    {
+        Write-Output "$Name version $Version is already installed!"
+        $doInstall = $false
+    }
+}
+        
+if ($doInstall)
+{
+    dotnet tool install $Name --version $Version $scopeString
+}
```
sarahelsaig: Please include an empty line at the end.

.github/actions/verify-dotnet-consolidation/action.yml
sarahelsaig: Since Install-DotNetTool has a wider use-case it could be a good idea to turn it into a standalone action. That way it could be called even externally, or re-used in other actions of our own without referring to its physical location. (also using GHA parameters is nicer) What do you think?

Piedone: Good idea!

Scripts/Install-DotNetTool.ps1
BenedekFarkas: Nope, regardless of downgrading or upgrading.

Scripts/Install-DotNetTool.ps1
BenedekFarkas: Local is the default behavior, and the wrapper shouldn't change that.

.github/actions/verify-dotnet-consolidation/action.yml
BenedekFarkas: I thought about that but wanted to get it working first. I'll test some further changes and get down to that.

Scripts/Install-DotNetTool.ps1
Piedone: Why though? This is our script, primarily we'll use it, and in the context of GitHub Actions, where we only need global.

BenedekFarkas: I find it frustrating when a wrapper changes the behavior compared to what it's wrapping when there's a corresponding configuration option (even if it was the other way around, i.e., $Local switch), I think that's bad DX.
However, since we'll wrap it in its own action, so the context is surely GHA (not local development for example, where this script can still be useful), then I think it's OK there, because it needs to be --global.

Piedone: I see the point, and OK then.

.github/actions/install-dotnet-tool/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Install .NET tool
+description: >
+  Installs a .NET tool globally.
```
Piedone:
```suggestion
description: Installs a .NET tool globally.
```

Piedone (changes requested):

PR #74, opened 2022-10-21
OSOE-397: Migrating from deprecated set-output commands

### LINKED ISSUE #68: Migrate from deprecated set-output commands (OSOE-397)
The `Write-Output "::set-output name=my-output::$myOutput"`-style commands [are now deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). Follow the instructions in the announcement to migrate away from them.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-397)

MZole:
[OSOE-397](https://lombiq.atlassian.net/browse/OSOE-397)

Fixes #68


.github/actions/build-dotnet/Write-CacheConfiguration.ps1
```diff
@@ -1,6 +1,6 @@
 param($IsNuget, $IsNpm, $RestoreKeys, $Hash)
 
-function Set-Output($Key, $Value) { Write-Output "::set-output name=$Key::$Value" }
```
Piedone, line 3: We actually have a function for this? Great, please then refactor the rest of the code to use this one instead of the custom `Write-Output`s. Such common scripts are under the root Scripts folder, you can see how those are accessed.

.github/actions/build-dotnet/Write-CacheConfiguration.ps1
```diff
@@ -1,6 +1,6 @@
...
+function Set-Output($Key, $Value) { Write-Output "$Key=$Value" >> $GITHUB_OUTPUT }
```
Piedone: Isn't this enough?

```suggestion
function Set-Output($Key, $Value) { "$Key=$Value" >> $GITHUB_OUTPUT }
```

.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -89,12 +89,12 @@ jobs:
         - ${{ inputs.timeout-minutes }}
     timeout-minutes: ${{ matrix.timeout-minutes }}
     steps:
-    - uses: actions/checkout@v3
```
Piedone, line 106: Are all these warnings coming from Test Reports? https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3362208732 Please run a build with just that step removed to see if all vanish.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -145,7 +145,7 @@ jobs:
       id: set-up-runtime-option
       if: ${{ inputs.runtime != '' }}
       run: |
-        Write-Output "::set-output name=runtime-option::--runtime=${{ inputs.runtime }}"
+        Set-Output "runtime-option=--runtime" "${{ inputs.runtime }}"
```
Piedone: The equality sign shouldn't be part of the `Set-Output` call, and the parameters don't match the original one.

.github/actions/check-merge-conflict/action.yml
```diff
@@ -7,7 +7,7 @@ runs:
     - name: Check for merge conflict
       if: github.event.pull_request == '' || github.event.pull_request.head.repo.fork == false
       # releases/2.x
-      uses: eps1lon/actions-label-merge-conflict@b8bf8341285ec9a4567d4318ba474fee998a6919
+      uses: eps1lon/actions-label-merge-conflict@v2.1.0
```
Piedone: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=For%20non%2DGitHub%20actions

Scripts/Set-Output.ps1
```diff
@@ -0,0 +1,11 @@
...
+    [Parameter(Mandatory = $true, Position = 0)]
+    [string]
+    $Key,
+
+    [Parameter(Mandatory = $true, Position = 1)]
+    [string]
+    $Value
+)
+
+return Write-Output "$Key=$Value" >> $GITHUB_OUTPUT
```
Piedone: This is duplicated with Write-CacheConfiguration.ps1. Also, check all affected workflows: Is this script really available in each of them? Because it's not loaded.

Piedone (changes requested):

.github/workflows/build-and-test-orchard-core.yml
MZole: Yes, you can see the results on [this run](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3364050166/jobs/5577951452), there are no warnings inside the `Tests`.

Scripts/Set-Output.ps1
MZole: The only workflows utilizing this are the `create-jira-issues`, `publish-nuget` and `deploy-to-azure`, I'm not sure if I can test it with `publish-nuget` or `deploy-to-azure`, but I will create a test issue with `create-jira-issues`  to check.

.github/workflows/build-and-test-orchard-core.yml
Piedone: This still has many warnings. The warnings show up on the Summary page, not under each job, like here at the bottom: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3364050166

Scripts/Set-Output.ps1
Piedone: You can make dummy runs of these workflows in each case (like you did with `create-jira-issues`). The full run doesn't need to succeed, but it has to pass the first point where this new script is used.

BTW unless you load the _Scripts_ folder at the beginning of the action (not workflow) then calls to this script won't find it. See e.g.: https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/create-jira-issues-for-community-activities/action.yml#L25

.github/workflows/build-and-test-orchard-core.yml
MZole: They are from the non-nuget builds, because they are using the dev branch of the workflow, as you requested [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/241#discussion_r1009726869). In every build the Tests step generates 6 warnings, that's why we had 24, and on this run, we only have 12. So if I would change back the branch to my issue branch on the root solution builds, there would be 0 warnings.

Piedone: OK, then please demonstrate this with using the issue branch for just this build.

Scripts/Set-Output.ps1
MZole: Managed to make this function work [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3365226333/jobs/5580478419) finally. The gaijira login wouldn't start if the set-output function wouldn't work.

.github/workflows/build-and-test-orchard-core.yml
MZole: [Here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3365284817) is the run.

Scripts/Set-Output.ps1
Piedone: Great!

.github/workflows/build-and-test-orchard-core.yml
Piedone: Great, thank you.

.github/actions/test-dotnet/action.yml
```diff
@@ -68,7 +68,7 @@ runs:
           ${{ inputs.build-directory }}/FailureDumps/
           test.out
         if-no-files-found: ignore
-
+
```
Piedone: Revert this whitespace change.

.github/actions/build-dotnet/Write-CacheConfiguration.ps1
```diff
@@ -19,8 +18,8 @@ $paths | % { New-Item -ItemType Directory -Force $_ } | Out-Null
...
 # the format function like this.
-Set-Output 'paths' ($paths -join '{0}')
+Set-Output('paths' ,($paths -join '{0}'))
 
-Set-Output 'cache-enabled' 'true'
-Set-Output 'key' "${RestoreKeys}-${Hash}"
-Set-Output 'restore-keys' $RestoreKeys
+Set-Output('cache-enabled', 'true')
+Set-Output('key', "${RestoreKeys}-${Hash}")
+Set-Output('restore-keys', $RestoreKeys)
```
Piedone: Why are the brackets and commas needed? `Set-Output` is a script, not a function.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -33,10 +33,10 @@ runs:
       shell: pwsh
       run: |
         $output = "${{ github.event.pull_request }}" -eq "" ? "False" : (Confirm-PullRequestTitle "$($Env:PR_TITLE)")
-        Set-Output "is-issue-pr" $output
+        Set-Output -Key "is-issue-pr" -Value $output
```
Piedone: Are you sure positional parameters are not working? Same in .github/actions/publish-nuget/action.yml. Now we have three kinds of syntaxes being used for `Set-Output` throughout the repo.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -33,10 +33,10 @@ runs:
...
 
     - name: Login to Jira
-      if: steps.issue-related-pr.outputs.is-issue-pr == 'False'
+      if: ${{ steps.issue-related-pr.outputs.is-issue-pr }} == 'False'
```
Piedone: These changes, as well as those below, shouldn't be needed. Same in .github/actions/publish-nuget/action.yml.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -84,10 +84,10 @@ runs:
 
         $fields += '}'
 
-        Set-Output "fields" $fields
+        Set-Output -Key "fields" -Value $fields
 
     - name: Create Jira Issue
-      if: steps.issue-related-pr.outputs.is-issue-pr == 'False'
+      if: ${{ steps.issue-related-pr.outputs.is-issue-pr }} 'False'
```
Piedone: With the equality sign removed this expression is invalid.

Scripts/Set-Output.ps1
```diff
@@ -8,4 +8,4 @@ param (
     $Value
 )
 
-return Write-Output "$Key=$Value" >> $GITHUB_OUTPUT
+Write-Output $Key=$Value >> $GITHUB_OUTPUT
```
Piedone: Are you sure the quotes are not needed around `$Key=$Value`? Also, is `Write-Output` really needed, isn't just writing to the variable enough?

Scripts/Set-Output.ps1
Piedone: I still don't see the Scripts folder being loaded for all affected actions.

Piedone (changes requested):

.github/actions/create-jira-issues-for-community-activities/action.yml
MZole: If I remember correctly, it was only working with this change, but I will check to make sure.

Scripts/Set-Output.ps1
MZole: Yes, It is working as expected without the quotes, and I will check on the `Write-Output`.

MZole: Yup, the `Write-Output` is unnecessary.

.github/actions/create-jira-issues-for-community-activities/action.yml
MZole: I checked, and this is only working with this change.

Piedone: This would go against the logic of these expressions (either the whole expression needs to be wrapped into `${{ }}`, or rather, the `, if`s in particular, don't need this at all). So, please check this with a minimal repro: A workflow that in one step writes "False" to the output, and then checks in the next one. Try what happens if you change the output to "True".

MZole: The root of the problem was that I was using `$GITHUB_OUTPUT` in the script but it should have been `$Env:GITHUB_OUTPUT ` :dizzy_face:
Now it is working with the original if cases.

Piedone: Indeed!

.github/actions/publish-nuget/action.yml
```diff
@@ -31,6 +31,12 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Setup Scripts
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
```
Piedone, line 37: This is duplicated below.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -126,7 +126,13 @@ jobs:
         - ${{ inputs.timeout-minutes }}
     timeout-minutes: ${{ matrix.timeout-minutes }}
     steps:
-    - uses: actions/checkout@v3
+    - name: Setup Scripts
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+        (Resolve-Path "${{ github.action_path }}/../../../Scripts").Path >> $Env:GITHUB_PATH
```
Piedone: `action_path` won't work in a workflow (as opposed to an action). But you needn't refactor the `set-up-runtime-option` to an action for this; instead, the step's output can be passed around in an environment variable.

Piedone (changes requested):
See above.

.github/actions/publish-nuget/action.yml
MZole: Just making sure, do you want me to still include the scripts in this new step, or include them at the beginning of the original `Setup` step?

Piedone: I think having just Setup Scripts like this here is better. But then also rename the below Setup step to something more describing.

.github/actions/publish-nuget/action.yml
```diff
@@ -34,10 +34,9 @@ runs:
     - name: Setup Scripts
       shell: pwsh
       run: |
-        "${{ github.action_path }}" >> $Env:GITHUB_PATH
```
Piedone: As I've written, this step was good. It's good to have a separate step just to set up scripts, including for the action.

Piedone (changes requested):

.github/actions/publish-nuget/action.yml
Piedone: Fixed this.

PR #83, opened 2022-10-27
OSOE-420: Make Azurite and SQL Server installation configurable in Lombiq.GitHub.Actions

### LINKED ISSUE #82: Make Azurite and SQL Server installation configurable (OSOE-420)
...and in the OSOCE NuGet build we should leave them off.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-420)

Piedone:
OSOE-420
Fixes #82

### DIFF
--- a/.github/workflows/build-and-test-orchard-core.yml
+++ b/.github/workflows/build-and-test-orchard-core.yml
@@ -59,14 +59,14 @@ on:
       set-up-sql-server:
         required: false
         type: string
-        default: "true"
+        default: "false"
         description: >
           If set to "true", SQL Server Express will be installed and started. This is necessary if the app's UI tests
           connect to an SQL Server database.
       set-up-azurite:
         required: false
         type: string
-        default: "true"
+        default: "false"
         description: >
           If set to "true", the Azurite Azure Blob Storage emulator will be installed and started. This is necessary if
           the app's UI tests connect to Azure Blob Storage for file storage.
--- a/.github/workflows/build-and-test-orchard-core.yml
+++ b/.github/workflows/build-and-test-orchard-core.yml
@@ -62,7 +62,7 @@ on:
         default: "false"
         description: >
           If set to "true", SQL Server Express will be installed and started. This is necessary if the app's UI tests
-          connect to an SQL Server database.
+          connect to a SQL Server database.
       set-up-azurite:
         required: false
         type: string
--- a/.github/workflows/build-and-test-orchard-core.yml
+++ b/.github/workflows/build-and-test-orchard-core.yml
@@ -132,11 +132,11 @@ jobs:
         cache-version: ${{ inputs.build-cache-version }}
 
     - name: Set up SQL Server
-      if: set-up-sql-server == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -56,6 +56,20 @@ on:
           (https://github.community/t/how-to-clear-cache-in-github-actions/129038/5) until the associated issue
           (https://github.com/actions/cache/issues/2) is resolved and the feature is available not just via the RESTful
           API.
+      set-up-sql-server:
```
BenedekFarkas, line 59: I'd rather prefix these with "setup" instead of "set-up" to match the spelling of the action they're calling (but I also find it a bit weird how the usage of this verb changed, so you can't really separate the verb from the noun anymore).

.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -56,6 +56,20 @@ on:
...
+        required: false
+        type: string
+        default: "true"
+        description: >
+          If set to "true", SQL Server Express will be installed and started. This is necessary if the app's UI tests
+          connect to an SQL Server database.
```
0liver: Nitpicker's corner: I would prefer "**a SQL Server**", because it's very common to pronounce SQL Server as "sequel server".

Here's some very interesting historical background: https://learnsql.com/blog/sql-or-sequel/

For me the essence is, that MS SQL Server is pronounced "Em Es Sequel Server" (like Bill Gates does) while the language SQL is pronounced "ES Que EL".

I've opened a [Teams thread](https://teams.microsoft.com/l/message/19:525a1a3656c845ae92c5e2c0d3617cc7@thread.skype/1666859417402?tenantId=959ede5a-38e9-44e6-adc8-85553beb8220&groupId=d6a5d775-6fb9-41dc-a7bb-481d4f884046&parentMessageId=1666859417402&teamName=Lombiq%20Developers&channelName=General&createdTime=1666859417402&allowXTenantAccess=false) just for that 😀

.github/workflows/build-and-test-orchard-core.yml
0liver: Let's be strong and **not** surrender to broken grammar! 🤺

Go **set up** that SQL Server!

BenedekFarkas: I usually think about this kind of stuff the same way as you do, but language is fluid, so what may start out as being broken might become the norm over time.

0liver: You're right. But I think in this case, it's just an oversight, and I wouldn't carry it over to our own code base.

Piedone: The actions are only called "setup" not "set-up" to match the official "setup" actions of GitHub (e.g. `setup-dotnet`, `setup-node`). I'd keep these parameters though.

.github/workflows/build-and-test-orchard-core.yml
Piedone: Fine :D.

PR #89, opened 2022-11-04
OSOE-424: Test execution uses better artifact/test report name suffixes, and support test projects not under the "tests" folder

Piedone:
OSOE-424


.github/actions/test-dotnet/Merge-FailureDumps.ps1
```diff
@@ -1,7 +1,10 @@
-﻿param ($Directory)
+param ($Directory)
 
 New-Item -Type Directory "$Directory/FailureDumps"
-Get-ChildItem "$Directory/test" -Recurse |
+
+$rootDirectory = (Test-Path -Path "$Directory/test") ? "$Directory/test" : $Directory
```
dministro:
```suggestion
$testDirectory = "$Directory/test"
$rootDirectory = (Test-Path -Path $testDirectory) ? $testDirectory : $Directory
```

Why this check is required?
Can it be a parameter eg.: `$TestDirSuffix` with default value `test`?

dministro (changes requested):

.github/actions/test-dotnet/Merge-FailureDumps.ps1
Piedone: When `test-dotnet` is used as part of `build-and-test-orchard-core` on a standard conventional (Lombiq) OC solution, we use this action to run all the tests in all test projects that are in the _tests_ folder. However, when any of the following criteria are not true, the _tests_ subfolder won't work:

1. We run `test-dotnet` on a single selected project.
2. Tests are not in the _tests_ folder.

The folder in which the tests reside could be configurable, but I figured that the majority of OC solutions, not even just ours, follow the _tests_ folder convention. So the rest of the use-cases will primarily be nr. 1, like it is now fo https://github.com/OrchardCMS/OrchardCore/pull/11194. This supports that easily, without the necessity for additional configuration and without a performance hit either (unless you count the folder existence check; though even that was there before in the form of a `hashFiles()` call).

dministro (approved):

PR #91, opened 2022-11-05
NEST-338: Re-implement Azure Reset build with GitHub Actions

DemeSzabolcs:
NEST-338


.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
+name: Reset Environment
```
Piedone: Add docs in Workflows.md.

Piedone: Also change the file name.

```suggestion
name: Reset Azure Environment
```

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        required: false
+        type: string
+        default: Production
+        description: The slot name of the web app you want to reset the destination web app with. What you see at the
+          top of the blade on the Azure Portal, when you open the slot, before the app name in parenthesis.
+      resource-group-name:
+        required: true
+        type: string
+        description: Name of the resource group.
+      url:
```
Piedone: Clarify if source or destination.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        type: string
+        description: The name of the database connection string.
+      database-master-connection-string-name:
+        required: true
+        type: string
+        description: The name of the database master connection string.
+      storage-connection-string-name:
+        required: true
+        type: string
+        description: The name of the storage connection string.
```
Piedone: Clarify what "master" is, and the name of what "name" refers to.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        description: Configuration for the timeout-minutes parameter of the workflow. The 360 is GitHub's default.
+      app-name:
+        required: true
+        type: string
+        description: What you see at the top of the blade on the Azure Portal. Can contain uppercase letters too.
+      destination-slot-name:
+        required: false
+        type: string
+        default: Staging
+        description: The slot name of the web app you want to reset. What you see at the top of the blade on the Azure
```
Piedone: This needs to use the `>` notation. Also below.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        required: true
+        type: string
+        description: The name of the storage connection string.
+      service-objective-name:
+        required: false
+        type: string
+        default: S0
+        description: The name of the service objective to scale the database to, e.g. B, S1, S4, P4, P6.
+      # These need to be stringified JSON because inputs don't support arrays, see 
+      # https://github.community/t/reusable-workflow-with-strategy-matrix/205676/2.
```
Piedone: It's not a JSON array.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+      container-exclude-list:
```
Piedone: What kind of container? Also below.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        required: false
+        type: string
+        default: "@()"
+        description: >
+            PowerShell string array with the name of the excluded container(s), e.g.
+            "@(\"not-media\")" or "@(\"not-media\", \"bad-container\")". The parameter must be a PowerShell string
+            array.
+      container-white-list:
```
Piedone: The opposite of an exclude list is an include list. Same for all the other cases here too.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        required: false
+        type: string
+        default: "@(\"media\")"
+        description: >
+            PowerShell string array with the name of the included container(s), e.g.
```
Piedone: Why do we need both an exclude and include list? What happens if one specifies non, just one, both? Explain in the description. Same for the folder config.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
+name: Reset Environment
+
+on:
+  workflow_call:
+    secrets:
+      AZURE_APP_SERVICE_RESET_SERVICE_PRINCIPAL:
+        required: true
+    inputs:
+      # This needs to be stringified JSON because inputs don't support arrays, see
```
Piedone: Does it make sense to run this on multiple machines at the same time with the same parameters? Because it doesn't seem so, and if not, then this workflow doesn't need to support a matrix. Same with `deploy-to-azure-app-service`.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -187,7 +190,10 @@ jobs:
       uses: Azure/powershell@38b0aa56c413239bd9dcc898d1f62986233658bc
       with:
         inlineScript: |
-          Stop-AzWebAppSlot -ResourceGroupName ${{ inputs.resource-group-name }} -Name ${{ inputs.app-name }} -Slot ${{ inputs.slot-name }}
+          Stop-AzureWebAppSlot `
```
Piedone: This and `Start-AzureWebAppSlot` below are custom scripts, right? Because then these steps should rather just use `run` with `pwsh`. Same in `reset-environment`, also for all the other steps.

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+        inlineScript: |
+          Set-AzureWebAppSqlDatabaseServiceObjective `
+            -ResourceGroupName ${{ inputs.resource-group-name }} `
+            -WebAppName ${{ inputs.app-name }} `
+            -SlotName ${{ inputs.destination-slot-name }} `
+            -ConnectionStringName ${{ inputs.database-master-connection-string-name }} `
+            -ServiceObjectiveName ${{ inputs.service-objective-name }}
+        azPSVersion: latest
+
+    - name: Add destination contained User to destination database
```
Piedone: Below too.

```suggestion
    - name: Add Destination Contained User to Destination Database
```

.github/workflows/reset-environment.yml
```diff
@@ -0,0 +1,219 @@
...
+      with:
+        inlineScript: |
+          Start-AzureWebAppSlot `
+            -ResourceGroupName ${{ inputs.resource-group-name }} `
+            -WebAppName ${{ inputs.app-name }} `
+            -SlotName ${{ inputs.destination-slot-name }}
+        azPSVersion: latest
+
+    - name: Ping Web App Slot
+      run: curl --connect-timeout 100 --retry 3 --retry-delay 15 ${{ inputs.url }}
```
Piedone: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=Prefer%20to%20use%20PowerShell%207%2B%20scripts Same for `deploy-to-azure-app-service`.

Piedone (changes requested):

.github/workflows/deploy-to-azure-app-service.yml
DemeSzabolcs: Yes they are custom scripts, but if I use the default `run` and `pwsh` then I need to add an extra step, see: https://lombiq.atlassian.net/browse/NEST-338?focusedCommentId=71347

Piedone: If you mean that "Install and import modules" then isn't it supposed to happen with _Lombiq/Infrastructure-Scripts/.github/actions/initialize_?

DemeSzabolcs: _Lombiq/Infrastructure-Scripts/.github/actions/initialize_ copies the modules, so they will be usable in this repository, but it doesn't installs/imports them. I will try to add that step to it.

.github/workflows/reset-environment.yml
DemeSzabolcs: Actually I think we have a custom script for this. I will use that.

Docs/Workflows.md
```diff
@@ -239,3 +239,28 @@ jobs:
       issue-component: Lombiq.MyProject
 
 ```
+
+## Reset Azure Environment workflow
+
+This workflow resets an Azure Environment, by replacing the Media Library and the Database with the ones from a given source slot. Example _reset-azure-environment.yml_:
```
```suggestion
This workflow resets an Azure Environment, by replacing the Orchard Core Media Library and the Database with the ones from a given source slot. Example _reset-azure-environment.yml_:
```

.github/workflows/reset-azure-environment.yml
```diff
@@ -0,0 +1,193 @@
...
+        required: true
+        type: string
+        description: Name of the resource group.
+      database-connection-string-name:
+        required: true
+        type: string
+        description: >
+          The name of the database connection string. The Connection String Name property specifies the name of the
+          named connection string that the SQL Workflow Instance Store should use to connect to the persistence
+          database.
```
Piedone: I.e. `database-connection-string-configuration-name`, what's configured as one of the connection strings under the app? I think I know what this refers to but even I'm confused by the description :).

Same for the other two `*connection-string-name`s.

.github/workflows/reset-azure-environment.yml
```diff
@@ -0,0 +1,193 @@
...
+          When valid values are provided for "blob-container-include-list", then "blob-container-exclude-list" is not
+          taken into consideration.
+      blob-container-include-list:
+        required: false
+        type: string
+        default: "@(\"media\")"
+        description: >
+          PowerShell string array with the name of the included Blob Container(s), e.g.
+          "@(\"media\")" or "@(\"media\", \"cool-container\")". The parameter must be a PowerShell string array. When
+          valid values are provided, it cancels out "blob-container-exclude-list".
```
Piedone: It's unclear what "cancels out" means. If the point is that if `blob-container-include-list` is valid then exclude-list won't be used, then repeated the description under the latter. Also explain what "valid" is, because the defaults seem valid too.

Same for folder.

.github/workflows/reset-azure-environment.yml
```diff
@@ -0,0 +1,193 @@
...
+      uses: Lombiq/Infrastructure-Scripts/.github/actions/initialize@issue/NEST-338
+
+    - name: Stop Web App Slot
+      run: |
+        Stop-AzureWebAppSlot `
+          -ResourceGroupName ${{ inputs.resource-group-name }} `
+          -WebAppName ${{ inputs.app-name }} `
+          -SlotName ${{ inputs.destination-slot-name }}
+
+    - name: Replace Media
```
Piedone, line 125: This too unexpectedly long time here: https://github.com/Lombiq/DotNest-Tenants-Core/actions/runs/3404808941/attempts/3 While attempt 2 was normal: https://github.com/Lombiq/DotNest-Tenants-Core/actions/runs/3404808941/attempts/2

Piedone (changes requested):

.github/workflows/reset-azure-environment.yml
DemeSzabolcs: Yes. I will use a description something like that. Btw I copied this one from the official [docs](https://learn.microsoft.com/en-us/dotnet/framework/windows-workflow-foundation/connection-string-and-connection-string-name). It's hard not to over-explain and underexplain it.

.github/workflows/reset-azure-environment.yml
DemeSzabolcs: It is because of the frequent runs. I tested it multiple times. When I ran it only 1-2 minutes after the previous finished, it always resulted in 20-30 (even 30+) minutes long Replace Media step. If you wait 10-15 minutes after a run it will average around 7-10 minutes for the replace media step.

Piedone: I see, that's OK then.

Piedone (approved):

PR #92, opened 2022-11-05
OSOE-392: Allowing new words in Spelling

MZole:
[OSOE-392](https://lombiq.atlassian.net/browse/OSOE-392)

### DIFF
--- a/.github/workflows/build-and-test-orchard-core.yml
+++ b/.github/workflows/build-and-test-orchard-core.yml
@@ -142,7 +142,7 @@ jobs:
         location: ${{ inputs.build-directory}}
 
     - name: Tests
-      uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@dev
+      uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@issue/OSOE-392
       with:
         build-directory: ${{ inputs.build-directory }}
         test-verbosity: ${{ inputs.build-verbosity }}
--- a/.github/workflows/build-and-test-orchard-core.yml
+++ b/.github/workflows/build-and-test-orchard-core.yml
@@ -142,7 +142,7 @@ jobs:
         location: ${{ inputs.build-directory}}
 
     - name: Tests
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -142,7 +142,7 @@ jobs:
         location: ${{ inputs.build-directory}}
 
     - name: Tests
-      uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@dev
+      uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@issue/OSOE-392
```
Piedone: Why is this necessary? There're no changes in `test-dotnet`.

Piedone (changes requested):

.github/workflows/build-and-test-orchard-core.yml
MZole: I have changed the connection string previously in `Invoke-SolutionTests`, but it was merged in OSOE-403, that's why the branch is changed, but yes, this is unnecessary now.

Docs/Workflows.md
```diff
@@ -71,6 +71,7 @@ There are more configuration files available, for more information visit the act
 You can provide these files in your own repository, under the path `.github/actions/spelling`. This can't be configured for another path.
 
 You can also use already existing configuration files by setting the `spell-check-this` parameter to another existing repository, where the files are found in the above mentioned path.
+This parameter is needed even if you want to update our dictionary, so changing just the workflow's branch from `dev` to your branch wont take any effect, you can leave it as it is.
```
```suggestion
You can also use already existing configuration files by setting the `spell-check-this` parameter to another existing repository, where the files are found in the above-mentioned path. This parameter is needed even if you want to update our dictionary in a custom branch of project consuming this workflow; changing just the workflow's branch from `dev` to your branch won't take any effect, you can leave it as it is.
```

Piedone (changes requested):

Piedone (approved):

PR #99, opened 2022-11-11
OSOE-466: Configuration option to filter tests

Piedone:
OSOE-466

### DIFF
--- a/.github/actions/test-dotnet/Invoke-SolutionTests.ps1
+++ b/.github/actions/test-dotnet/Invoke-SolutionTests.ps1
@@ -44,8 +44,8 @@ foreach ($test in $tests) {
         '--nologo',
         '--logger', 'trx;LogFileName=test-results.trx'
         '--verbosity', $Verbosity
-        $Filter ? "" : "--filter"
-        $Filter ? "" : $Filter
+        $Filter ? '' : '--filter'
+        $Filter ? '' : $Filter
         $test
     )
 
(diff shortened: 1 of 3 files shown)


.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -35,7 +37,19 @@ $tests = dotnet sln list |
...
-    dotnet test -c Release --no-restore --no-build --nologo --logger "trx;LogFileName=test-results.trx" --verbosity $Verbosity $test 2>&1 >test.out
+    $dotnetTestSwitches = @(
+        '--configuration', 'Release'
+        '--no-restore',
+        '--no-build',
+        '--nologo',
+        '--logger', 'trx;LogFileName=test-results.trx'
+        '--verbosity', $Verbosity
+        [string]::IsNullOrEmpty($Filter) ? '' : "--filter"
+        [string]::IsNullOrEmpty($Filter) ? '' : $Filter
```
sarahelsaig: `[string]::IsNullOrEmpty` is redundant, also the string quotes are inconsistent.
```suggestion
        $Filter ? "" : "--filter"
        $Filter ? "" : $Filter
```

sarahelsaig (changes requested):

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
Piedone: All of them should actually be single quote (since we don't need string interpolation).

Piedone: It's not really working as expected:

image

I'm reverting to the original code

sarahelsaig: Yeah, the ternary's components should be the other way around, since "truthy" implies that it's not null or empty.

Piedone: Eh.

PR #100, opened 2022-11-11
OSOE-433: Implement workflow cancelation for all jobs when given workflow/job fails

### LINKED ISSUE #94: Implement workflow cancelation from independent jobs, and recommend it for Spelling workflows (OSOE-433)
If multiple independent jobs run in the same workflow, then [they can't be configured](https://github.com/orgs/community/discussions/38361) to cancel the whole run if they fail. This is an issue in the likes of [OSOCE `build-and-test`](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/blob/dev/.github/workflows/build-and-test.yml) because the Spelling workflow will finish quickly and may fail, but the rest of the workflow will continue to run, though should rather be canceled. While dependent jobs solve this, that would remove parallelization.

Let's implement this feature in a reusable way. Ideally, it can be done in a way that doesn't require all reusable workflows to support it (but this may rather be possible from a separate job). If not, then add this as an optional, disabled-by-default feature to all of our reusable workflows.

- [Document this](https://github.com/Lombiq/GitHub-Actions/blob/dev/Docs/Workflows.md#spelling-workflow) as the recommended setup for Spelling when used as an independent job in a larger workflow.
- Configure OSOCE `build-and-test` to utilize this with Spelling.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-433)

Piedone:
OSOE-433
Fixes #94.

### DIFF
--- a/.github/actions/cancel-workflow/action.yml
+++ b/.github/actions/cancel-workflow/action.yml
@@ -1,0 +1,14 @@
+name: Cancel Workflow
+description: >
+  Cancels the current workflow run, i.e. all jobs. Useful if you want to cancel the rest of the workflow when one job
+  fails.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Cancel Workflow
+      shell: pwsh
+      run: |
+        $url = "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel"
+        $headers = @{"Authorization" = "Bearer ${{ env.GITHUB_TOKEN }}"; "Accept" = "application/vnd.github+json"}
+        Invoke-WebRequest $url -Headers $headers -Method Post
--- a/.github/actions/cancel-workflow/action.yml
+++ b/.github/actions/cancel-workflow/action.yml
@@ -7,6 +7,8 @@ runs:
   using: "composite"
   steps:
     - name: Cancel Workflow
+      # Fork PRs won't have a token with write access to Actions, thus won't be able to cancel the workflow.
+      if: github.event.pull_request == '' || github.event.pull_request.head.repo.fork == false
       shell: pwsh
       run: |
         $url = "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel"
--- a/.github/actions/cancel-workflow/action.yml
+++ b/.github/actions/cancel-workflow/action.yml
@@ -10,5 +10,5 @@ runs:
       shell: pwsh
       run: |
         $url = "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel"
-        $headers = @{"Authorization" = "Bearer ${{ env.GITHUB_TOKEN }}"; "Accept" = "application/vnd.github+json"}
+        $headers = @{"Authorization" = "Bearer ${{ secrets.GITHUB_TOKEN }}"; "Accept" = "application/vnd.github+json"}
         Invoke-WebRequest $url -Headers $headers -Method Post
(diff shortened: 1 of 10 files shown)


.github/actions/cancel-workflow/action.yml
```diff
@@ -0,0 +1,16 @@
...
+  using: "composite"
+  steps:
+    - name: Cancel Workflow
+      # Fork PRs won't have a token with write access to Actions, thus won't be able to cancel the workflow.
+      if: github.event.pull_request == '' || github.event.pull_request.head.repo.fork == false
+      shell: pwsh
+      run: |
+        $url = "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel"
+        $headers = @{"Authorization" = "Bearer ${{ env.GITHUB_TOKEN }}"; "Accept" = "application/vnd.github+json"}
+        Invoke-WebRequest $url -Headers $headers -Method Post
```
dministro, line 16: Great idea!

.github/workflows/spelling.yml
```diff
@@ -62,3 +70,9 @@ jobs:
           extra-dictionaries: ${{ inputs.extra-dictionaries }}
           config: ${{ inputs.config }}
           spell-check-this: ${{ inputs.spell-check-this }}
+
+      - name: Cancel Workflow on Failure
```
dministro, line 74: > [Document this](https://github.com/Lombiq/GitHub-Actions/blob/dev/Docs/Workflows.md#spelling-workflow) as the recommended setup for Spelling when used as an independent job in a larger workflow.

was not addressed.

I see the `cancel-workflow` action and the `cancel-workflow-on-failure` are well documented, but noting this behavior under not just the `spelling` but all affected workflow documentation would be necessary.

.github/actions/cancel-workflow/action.yml
```diff
@@ -0,0 +1,16 @@
...
+description: >
+  Cancels the current workflow run, i.e. all jobs. Useful if you want to cancel the rest of the workflow when one job
+  fails.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Cancel Workflow
+      # Fork PRs won't have a token with write access to Actions, thus won't be able to cancel the workflow.
+      if: github.event.pull_request == '' || github.event.pull_request.head.repo.fork == false
```
dministro, line 11: We use this condition in `check-merge-conflict` workflow too. Can't we DRY this?
I mean we could add a `setup` step and with `Set-Output` we can add eg.: `has-write-access` (until no official solution for this) and check the `has-write-access` here and in `check-merge-conflict`.

What do you think about this?

dministro (changes requested):

.github/workflows/spelling.yml
Piedone: I didn't do that because while initially, I thought this will be an optional setup for just Spelling that you as a consumer of these workflows should do, but then this became the default behavior for most workflows. But a mention of this under workflows is indeed necessary.

.github/actions/cancel-workflow/action.yml
Piedone: Thanks!

.github/actions/cancel-workflow/action.yml
Piedone: I'd love to DRY this but we'd have the opposite effect. For this, we'd need a composite action and call it like this:

```yml
- name: Check Fork PR
  id: check-fork-pr
  uses: Lombiq/GitHub-Actions/.github/actions/check-fork-pr@dev

- name: Cancel Workflow
  if: steps.check-fork-pr.outputs.is-fork-pr != 'true'
...
```

This duplicates the amount of code we'd need to copy-paste around (plus the action itself). While granted, the logic and docs, which matter more, would be centralized (to an extent; the docs are slightly different due to the different operations), this still seems like a bad deal to me.

dministro (approved):

PR #101, opened 2022-11-12
OSOE-464: Making dotnet test output go to the build output to aid debugging, as well as printing ITestOutputHelper content

Piedone:
OSOE-464


.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -35,21 +37,27 @@ $tests = dotnet sln list |
...
     }
 
-    $needsGrouping = (Select-String "::group::" test.out).Length -eq 0
-
-    if ($needsGrouping) { Write-Output "::group::Test Failed: $test" }
-
-    bash -c "cat test.out | grep -v 'Connection refused \[::ffff:127.0.0.1\]' | grep -v 'ChromeDriver was started successfully'"
-
-    if ($needsGrouping) { Write-Output "::endgroup::" }
-
```
Piedone: Note that despite this missing, there's per-test grouping in the build output, see e.g. [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3452657800/jobs/5762708404#step:9:80):

image

sarahelsaig: per-test grouping comes from the UI Testing Toolbox [here](https://github.com/Lombiq/UI-Testing-Toolbox/blob/5d070b36e0cbe99a8bf92aac38f0558b5a876185/Lombiq.Tests.UI/Services/GitHubActionsGroupingTestOutputHelper.cs#L40).
The downside is that now the per-process logs around the tests are still not grouped and when a test succeeds the logs are still printed, polluting the the GHA output.

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -43,27 +43,21 @@ foreach ($test in $tests) {
         '--no-build',
         '--nologo',
         '--logger', 'trx;LogFileName=test-results.trx'
+        # This is for xUnit ITestOutputHelper, see https://xunit.net/docs/capturing-output.
+        '--logger', 'console;verbosity=detailed'
         '--verbosity', $Verbosity
```
sarahelsaig, line 57: Doesn't that override or conflict with the `'--verbosity', $Verbosity` below?

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -43,27 +43,21 @@ foreach ($test in $tests) {
...
     {
         Write-Output "Test Successful: $test"
         continue
     }
 
-    $needsGrouping = (Select-String "::group::" test.out).Length -eq 0
-
-    if ($needsGrouping) { Write-Output "::group::Test Failed: $test" }
-
-    bash -c "cat test.out | grep -v 'Connection refused \[::ffff:127.0.0.1\]' | grep -v 'ChromeDriver was started successfully'"
```
sarahelsaig, line 64: Could we still filter out these pointless "Connection refused" and "ChromeDriver" log lines on `dotnet test @dotnetTestSwitches`?

sarahelsaig (changes requested):

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
Piedone: No, since this verbosity is only for the given logger (you can set multiple loggers, and each with their own verbosity).

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
Piedone: Printing a succeeding test's log as well is desired and one of the goals with these changes. We need to be able to compare successful and failing tests. But right, I'll readd grouping for the processes too.

Piedone: Ah well, there are no nested groups... https://github.com/actions/runner/issues/1477

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
Piedone: I don't know or have found a way to live-filter a streaming output, though I only have smallish experience with PS.

The goal here is for the output of the tests to show up in the log continuously (to pinpoint what the last thing was before the runner was killed due to timeout). If there's a file at the end of the `dotnet test` call then you can grep like this, but we can't wait for a file.

We could still redirect to a file and tail it and that [can be filtered](https://stackoverflow.com/a/48773243/220230) but going through a file doesn't seem like the best idea, and we also need to somehow know when the stream actually ended and quit the `Wait`.

So, I don't know. It seems to me that implementing grouping around the test project if it doesn't have its custom group and letting UI tests group themselves, as well as filtering out such messages need a file. If we have a file, however, we won't have a continuous output.

Do you have any suggestions by chance?

BTW I'm also trying to figure out where the annoying log messages come from and disable them in the first place.

sarahelsaig: `dotnet test @dotnetTestSwitches 2>&1 | ? { $_ -notlike '*Connection refused [[]::ffff:127.0.0.1[]]*' -and $_ -notlike '*ChromeDriver was started successfully*' }` should filter it line-by-line as it's passed to the pipe. We can live without the outer grouping until they finally add named groups to the runner.

Piedone: Ah nice, so this works like this!

.github/actions/cancel-workflow/action.yml
```diff
@@ -1,7 +1,13 @@
...
 description: >
   Cancels the current workflow run, i.e. all jobs. Useful if you want to cancel the rest of the workflow when one job
-  fails.
+  fails. Note that this will cause the workflow to appear cancelled, not failed.
+
+# Cancelling the workflow in a post-script (like this:
+# https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runspost; can also be done with
+# this action: https://github.com/webiny/action-post-run, see Git history of this file) wouldn't help the status, it
+# would still be cancelled. It actually indeed is, but it would be nicer to set it to failed, but there seems to be no
+# way to do this.
```
Piedone, line 10: I had some fun experimenting around with this without avail.

But at least I had fun.

image

PR #106, opened 2022-11-17
CARE-265: Adding .NET Framework build features

Piedone:
[CARE-265](https://lombiq.atlassian.net/browse/CARE-265)


.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+    -p:Version=$Version
+    $Switches
+"@
+
+Write-Output "Building solution or project with ``msbuild $SolutionOrProject $($buildSwitches -join " ")``."
+
+msbuild $SolutionOrProject @buildSwitches
+
+# Without this, if the msbuild command fails with certain MSB error codes (not build errors), then still the wouldn't
+# cause this script to fail.
```
0liver:
```suggestion
# Without this, if the msbuild command fails with certain MSB error codes (not build errors), they still won't cause
# this script to fail.
```

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+# error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.1 were not found. To resolve this, install the
+# Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET
+# Framework Developer Packs at https://aka.ms/msbuild/developerpacks 
+
+if ($?)
+{
+    Write-Output "Build successful."
+    continue
```
0liver: Where does this jump to?

.github/actions/checkout/action.yml
```diff
@@ -0,0 +1,20 @@
+name: Checkout
+description: >
+  Runs actions/checkout. Only to centralize which version we use from the action. Intentionally not documented in
+  Actions.md since it's meant to be only for internal use.
```
0liver:
```suggestion
  Runs actions/checkout. Exists nly to centralize which version of the action we use. Intentionally not documented in
  Actions.md since it's only meant to be for internal use.
```

.github/actions/msbuild/action.yml
```diff
@@ -0,0 +1,73 @@
...
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+        (Resolve-Path "${{ github.action_path }}/../../../Scripts").Path >> $Env:GITHUB_PATH
+
+    - name: Add MSBuild to PATH
+      # v1.1.3
+      uses: microsoft/setup-msbuild@34cfbaee7f672c76950673338facd8a73f637506
+      with:
+        vs-version: "[17.3,]"
```
0liver, line 51: Did you intentionally not allow lower (minor) versions of VS?

.github/actions/msbuild/action.yml
```diff
@@ -0,0 +1,73 @@
...
+
+    - name: Build and Static Code Analysis
+      shell: pwsh
```
0liver, line 54: Seeing `shell: pwsh` the second time in this file, I'm wondering if that shouldn't simply be the default for the whole action.

.github/actions/msbuild/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    description: Verbosity parameter for msbuild.
+  enable-code-analysis:
+    required: false
+    default: "true"
+    description: If set to "true", static code analysis is enabled during the build.
+  publish-version:
+    required: false
+    default: "1.${{ github.run_number }}.${{ github.run_attempt }}-${{ github.run_id }}"
+    description: The value given to `msbuild`'s `-Version` switch.
+  dotnet-build-switches:
```
0liver: Seems to me that `msbuild-switches` would be the more appropriate name for this parameter.

.github/actions/msbuild/action.yml
```diff
@@ -0,0 +1,73 @@
...
+            SolutionOrProject = (Get-ChildItem ${{ inputs.solution-or-project-path }}).FullName
+            Verbosity = "${{ inputs.verbosity }}"
+            EnableCodeAnalysis = "${{ inputs.enable-code-analysis }}"
+            Version = "${{ inputs.publish-version }}"
+            Switches = $buildSwitches
+        }
+
+        $startTime = [DateTime]::Now
+        Build-DotNetFrameworkSolutionOrProject @switches
+        $endTime = [DateTime]::Now
```
0liver: For clarity and meaning, I would rename these two variables to `$givenSwitches` and `$allSwitches`:

```suggestion
        $givenSwitches = @'
            ${{ inputs.dotnet-build-switches }}
        '@

        $allSwitches = @{
            SolutionOrProject = (Get-ChildItem ${{ inputs.solution-or-project-path }}).FullName
            Verbosity = "${{ inputs.verbosity }}"
            EnableCodeAnalysis = "${{ inputs.enable-code-analysis }}"
            Version = "${{ inputs.publish-version }}"
            Switches = $givenSwitches
        }

        $startTime = [DateTime]::Now
        Build-DotNetFrameworkSolutionOrProject @allSwitches
        $endTime = [DateTime]::Now
```

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -26,13 +26,18 @@ else
 
 $Env:Lombiq_Tests_UI__BrowserConfiguration__Headless = "true"
 
+# Usually tests should be built with the Debug configuration. If that's selected, then most possibly the whole solution
+# was built with Release, so we need to build but not restore. Otherwise, the Release tests are already built, so we
+# don't need to build either.
```
0liver: Trying to make the thought process clearer:
```suggestion
# We assume that the solution was built in Release configuration. If the tests need to be built in Debug configuration,
# as they should, we need to first build them, but not restore. Otherwise, the Release tests are already built, so we
# don't need to build them here.
```

.github/workflows/msbuild-and-test.yml
```diff
@@ -0,0 +1,97 @@
+name: Build and Test Orchard Core solution
```
0liver: The name needs to be adjusted.

.github/workflows/msbuild-and-test.yml
```diff
@@ -0,0 +1,97 @@
...
+        required: false
+        type: string
+        default: "true"
+      machine-type:
+        required: false
+        type: string
+        default: "windows-latest"
+        description: >
+          The name of the type of the machine to run the workflow under, e.g. "windows-latest" or "windows-2022". Needs
+          to be Windows.
```
0liver: Simpler language:
```suggestion
          The machine type to run the workflow under, e.g. "windows-latest" or "windows-2022". Needs to be Windows.
```

.github/workflows/msbuild-and-test.yml
```diff
@@ -0,0 +1,97 @@
...
+        verbosity: ${{ inputs.build-verbosity }}
+        enable-code-analysis: ${{ inputs.build-enable-code-analysis }}
+
+    - name: Tests
+      uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@issue/CARE-265
+      with:
+        build-directory: ${{ inputs.build-directory }}
+        test-verbosity: ${{ inputs.build-verbosity }}
+        test-filter: ${{ inputs.test-filter }}
+        test-configuration: "Release"
```
0liver, line 90: Why did you choose `"Release"` here when we think that it should be `"Debug"`, usually?

.github/actions/build-dotnet/Build-DotNetSolutionOrProject.ps1
```diff
@@ -74,11 +74,7 @@ if ($noErrors -and !$?)
     exit 1
 }
 
-# With node reuse, dotnet build spawns processes that while speed up build, they can cause dotnet test and other dotnet
-# tools to randomly hang. So, here we shut down those processes for later actions.
-# For details see: https://github.com/Lombiq/UI-Testing-Toolbox/issues/228.
-Write-Output "Shutting down .NET build servers."
-dotnet build-server shutdown
+Close-DotNetBuildServers
```
0liver: How about:
```suggestion
Stop-DotNetBuildServers
```
?

0liver (changes requested):

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
Piedone: It's a copy-paste error, not needed.

.github/actions/msbuild/action.yml
Piedone: Yes. The current latest version of VS is 17.4.1, so this is the one we know for sure current projects work. That's not available in the GHA image, however, and the next best thing is 17.3 (while 17.x would most possibly work fine, there's no advantage to allowing that, since the runner images won't downgrade).

.github/actions/msbuild/action.yml
Piedone: While [you can set a default shell for a job](https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs#setting-default-shell-and-working-directory), you can't do that or anything similar within a composite action (which is a self-contained component). The [official docs](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file) do this awkward repetition too.

.github/actions/msbuild/action.yml
Piedone: This couldn't be committed directly here, so I manually committed.

.github/workflows/msbuild-and-test.yml
Piedone: `dotnet test` can only automatically build the project if it's a .NET (non-FW) one. So, to have Debug builds for these tests, we'd need to separately build them in this workflow. I don't see much point to that, since Debug is only useful for Shouldly, which in turn really only benefits from it in complex UI tests, something we don't have for .NET Framework.

.github/actions/checkout/action.yml
```diff
@@ -0,0 +1,20 @@
+name: Checkout
+description: >
+  Runs actions/checkout. Exists only to centralize which version of the action we use. Intentionally not documented in
+  Actions.md since it's only meant to be for internal use.
```
0liver:
```suggestion
  Actions.md since it's only meant for internal use.
```

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+    $rawInput.Replace("`r", "").Split("`n") | ForEach-Object { $PSItem.Trim() } | Where-Object { $PSItem }
+}
+
+Write-Output ".NET version number: $Version"
+
+if (Test-Path src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj)
+{
+    Write-Output "::group::Gulp Extensions found. It needs to be explicitly built before the solution."
+
+    # These need to be different than that of msbuild.
```
0liver:
```suggestion
    # These need to be different than those for msbuild.
```

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+    --verbosity:$Verbosity
+    --warnaserror
+    --warnAsMessage:MSB3026
+    --consoleLoggerParameters:NoSummary
+    -p:TreatWarningsAsErrors=true
+    -p:RunAnalyzersDuringBuild=$EnableCodeAnalysis
+    -p:Retries=4
+    -p:RetryDelayMilliseconds=1000
+    -p:Version=$Version
+"@
```
0liver: Let's indent that for better readability:
```suggestion
        --configuration:Release
        --nologo
        --verbosity:$Verbosity
        --warnaserror
        --warnAsMessage:MSB3026
        --consoleLoggerParameters:NoSummary
        -p:TreatWarningsAsErrors=true
        -p:RunAnalyzersDuringBuild=$EnableCodeAnalysis
        -p:Retries=4
        -p:RetryDelayMilliseconds=1000
        -p:Version=$Version
    "@
```

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+    $startTime = [DateTime]::Now
+    dotnet build src/Utilities/Lombiq.Gulp.Extensions/Lombiq.Gulp.Extensions.csproj @gulpBuildSwitches
+    $endTime = [DateTime]::Now
+
+    Write-Output ("Gulp Extensions build took {0:0.###} seconds." -f ($endTime - $startTime).TotalSeconds)
+    Write-Output "::endgroup::"
+}
+
+# - -p:Retries and -p:RetryDelayMilliseconds are to retry builds if it fails the first time due to random locks.
```
0liver: Remove the leading hyphen:
```suggestion
# -p:Retries and -p:RetryDelayMilliseconds are used to retry builds when they fail due to random locks.
```

.github/workflows/msbuild-and-test.yml
0liver: OK! I overlooked that you've already configured "Debug" as default [here](https://github.com/Lombiq/GitHub-Actions/pull/106/files#diff-24009ee87b154f96e015e85a7a1e66a585818e365a42e6a07d14872769a9b29aR23). 😁

0liver (changes requested):

.github/workflows/msbuild-and-test.yml
0liver: Still 😉

0liver (changes requested):

.github/workflows/msbuild-and-test.yml
Piedone: I guess you're looking at an older version because I fixed this: https://github.com/Lombiq/GitHub-Actions/commit/bc261f7f479c5847fca86b2b5c788be39c1fb40f

.github/actions/msbuild/Build-DotNetFrameworkSolutionOrProject.ps1
```diff
@@ -0,0 +1,77 @@
...
+        --verbosity:$Verbosity
+        --warnaserror
+        --warnAsMessage:MSB3026
+        --consoleLoggerParameters:NoSummary
+        -p:TreatWarningsAsErrors=true
+        -p:RunAnalyzersDuringBuild=$EnableCodeAnalysis
+        -p:Retries=4
+        -p:RetryDelayMilliseconds=1000
+        -p:Version=$Version
+"@
```
Piedone, line 32: The `@` string notation requires the at sign to be all to the left.

0liver (approved):

PR #111, opened 2022-11-22
OSOE-470: Remove false entries (actual typos) from spell checking allow lists in Lombiq.GitHub.Actions

porgabi:
[OSOE-470](https://lombiq.atlassian.net/browse/OSOE-470)


BenedekFarkas (changes requested):
Please check every entry that doesn't seem to be a proper word.

.github/actions/spelling/patterns.txt
```diff
@@ -19,3 +19,6 @@
 [Xx][Uu]nit
 [Xx][Pp][Aa][Tt][Hh]
 [Rr]eadme
+[{]?[0-9a-zA-Z]{8}-([0-9a-zA-Z]{4}-){3}[0-9a-zA-Z]{12}[}]?
+[s][h][a][0-9]{3}-[0-9a-zA-Z\+\/]{64}
+\b([a-f0-9]{40})\b
```
porgabi: Regular expressions for the following cases:
1. GUIDs with dashes.
2. CDN Integrity "sha" IDs, which do have a proper name that I forgot. Found in ResourceManagementOptions classes. For example:
image
3. Git commit hashes.

BenedekFarkas: Very nice!

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,9 @@
-[Ll][Oo][Mm][Bb][Ii][Qq]
```
0liver, line 1: The [documentation examples](https://github.com/check-spelling/check-spelling/wiki/Configuration-Examples%3A-patterns) actually use comments, which I think makes a lot of sense in combination with Regexes.

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,9 @@
...
-[Ss][Dd][Kk]
-[Rr][Ee][Aa][Dd][Mm][Ee]
-[Oo][Ss][Oo][Cc][Ee]
-[Ss][Ff][Cc]
+[Cc](loud|LOUD)[Ss](mith|MITH)
 [Xx][Uu]nit
-[Xx][Pp][Aa][Tt][Hh]
-[Rr]eadme
+[Xx][Pp]ath
+[{]?[0-9a-zA-Z]{8}-([0-9a-zA-Z]{4}-){3}[0-9a-zA-Z]{12}[}]?
```
0liver: Let's simply say that the GUID should be surrounded by non-word characters (colloquially speaking): 
```suggestion
\b[0-9a-zA-Z]{8}-([0-9a-zA-Z]{4}-){3}[0-9a-zA-Z]{12}\b
```

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,9 @@
...
+[s][h][a][0-9]{3}-[0-9a-zA-Z\+\/]{64}
```
0liver: This is equivalent to:
```suggestion
sha[0-9]{3}-[0-9a-zA-Z+/]{64}
```
No need to escape characters in character classes (enclosed by `[` and `]`).

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,9 @@
...
+\b([a-f0-9]{40})\b
```
0liver: Let's use the same order of digits and letters here as above:
```suggestion
\b([0-9a-f]{40})\b
```
Are uppercase letters not allowed here? If yes, then a comment should mention that.

.github/actions/spelling/patterns.txt
0liver: I'd have this to add:
```txt
# CSS colors - need the ( at the beginning to distinguish this from a comment.
(?:#[0-9a-fA-F]{3,6})\b
```

.github/actions/spelling/expect.txt
```diff
@@ -1,33 +1,27 @@
...
 Cms
 CDNs
 EACCES
 EULA
-Egyw
 GDPR
-Hsl
 IShortcode
-JIRA
+jira
```
0liver, line 12: Move to the lower-case section.

.github/actions/spelling/allow.txt
```diff
@@ -291,7 +278,7 @@ TValue
 Toolset
 Trumbowyg
 USERPROFILE
-Uncomment
+uncomment
```
0liver, line 283: Move down to the lower-case section.

0liver (changes requested):

0liver:
Btw, here's how to use those dictionaries from GHA:

```yaml
jobs:
  spelling:
    name: Spelling
    uses: Lombiq/GitHub-Actions/.github/workflows/spelling.yml@issue/CARE-275
    secrets:
      CHECKOUT_TOKEN: ${{ secrets.LOMBIQBOT_GITHUB_PERSONAL_ACCESS_TOKEN }}
    with:
      dictionary-source-prefixes: '{
        "cspell": "https://raw.githubusercontent.com/check-spelling/cspell-dicts/v20220816/dictionaries/",
        "gha": "https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/.github/actions/spelling/"
      }'
      extra-dictionaries:
        cspell:filetypes/filetypes.txt
        cspell:django/django.txt
        cspell:aws/aws.txt
        cspell:html/html.txt
        cspell:css/css.txt
        cspell:csharp/csharp.txt
        cspell:rust/rust.txt
        cspell:npm/npm.txt
        cspell:dotnet/dotnet.txt
        cspell:scala/scala.txt
        gha:allow.txt
        gha:excludes.txt
        gha:expect.txt
        gha:patterns.txt
```

0liver:
Looking forward to you merging this to avoid conflicts with my project issue.

.github/actions/spelling/expect.txt
BenedekFarkas: Alphabetical order is messed up completely, so we'll do that on dev after merge, not to spam the diff.

.github/actions/spelling/patterns.txt
BenedekFarkas: I think we can skip the `(`, because comments always (should) have a space after the `#`, but the hex color code is not always between parenthesis.

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,9 @@
-[Ll][Oo][Mm][Bb][Ii][Qq]
 (\((.*?)\))
```
BenedekFarkas: @Luko6 what's this pattern for?

.github/actions/spelling/patterns.txt
BenedekFarkas: @porgabi what about the uppercase letters here?

porgabi: Git commit hash IDs don't contain uppercase letters, so no need for those to be included in this one.

BenedekFarkas: Thanks!

.github/actions/spelling/patterns.txt
0liver: The parentheses incl. `?:` mark a non-capturing regex group, so they are not explicit characters.

I wasn't aware of the rule that comments need a space. If that's the case, that will also do, also the resemblance of the line to a comment is kind of awkward.

You decide.

.github/actions/spelling/patterns.txt
0liver: This will skip anything in parentheses, see https://regexr.com/73tll for an explanation.

Seems problematic to me.

BenedekFarkas: @Luko6 before we remove this, what was the intention here so that we can find a proper replacement?

Luko6: I think this is used to skip the links inside parentheses in markdown files

.github/actions/spelling/patterns.txt
BenedekFarkas: They usually do, although AFAIK it's difficult to enforce. I usually call it out in a review, but I think automatic code formatting fixes those too in case the space is missing.
I committed your regex modified slightly (https://github.com/Lombiq/GitHub-Actions/pull/111/commits/485d64a7c38f312e74f620538dbc447e32f3c4a9) - seems to work as expected.

.github/actions/spelling/patterns.txt
BenedekFarkas: OK, so the problem, like Oliver explained is that it's way too much unrestricted. Weren't the URL patterns sufficient for this use-case?

.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,22 @@
...
-[Ss][Ff][Cc]
+[Xx][Pp]ath
 [Xx][Uu]nit
-[Xx][Pp][Aa][Tt][Hh]
-[Rr]eadme
+
+(\((.*?)\))
+
+# URI/URL formats.
+(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})
```
0liver: I would put each pattern on a separate line. There's no need to OR them into a single, unreadable line:
```suggestion
https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}
www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}
https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}
www\.[a-zA-Z0-9]+\.[^\s]{2,}
```

Now it starts to get interesting! Because, e.g. a URL like `http://wwwombiq.com` will not match any of these.

Questions:
1. Why is `(?:www\.|(?!www))` in the regex? It means that www has to be followed by a dot OR not be at the beginning at all. That seems like an arbitrary constraint.
2. Why do we need 4 patterns in the first place? We could merge the 1st+3rd lines and 2+4 by allowing `-` to appear at the beginning or end of the domain name, which would be incorrect, but this is not a URL validator: `www\.[a-zA-Z0-9-]+\.[^\s]{2,}`

BenedekFarkas: @0liver I've committed your suggestion above, thanks!
1. I'm not sure, @Luko6 or @Piedone can answer maybe if there are edge cases we need to consider here.
2. I agree about simplification, even making a mistake like that would be detected with simple testing during development (i.e., does the link I pasted work?). Can you please rework these, based on your judgement, in a way that they are easy to read (and maintain, if necessary), but remain functional. Please also add examples in a comment what use case(s) they cover.

0liver: None of the above would allow localhost urls either.

Let's remember that this is only about spell-checking and ignoring string that look like URLs. So maybe simply:
```suggestion
https?:\/\/\S+
www.\S+
```
That will match strings starting with `http[s]://` or `www.` until the first whitespace character.

Piedone: I'm not sure what I can provide to the conversation about URL patterns :). Please look up a commonly used pattern (note that this should be perl regex here) and use that.

Also, note, that the patterns file is not about matching arbitrary strings, but lines: If a line matches the pattern, then the whole line will be excluded from spell-checking.

BenedekFarkas: I actually didn't know that patterns.txt ignores the whole line where it matches, that makes the recently removed "any text in parenthesis" pattern and the other, single-word entries even more problematic. I'll remove the rest of the single-word entries too, then - e.g., `[Hh]ome[Pp]age` in patterns can be substituted with `homepage` and `homePage` in allow.txt, but I'll test spelling without these entries to see which casing combinations we need to actually cover.
I think `https?:\/\/\S+` and `www.\S` are OK, although it would be better, if they didn't exclude the whole line from spellchecking, but simplifying them is already an improvement and I don't see an alternative.

PR #118, opened 2022-11-25
MENT-157: build-and-test-orchard-core workflow shouldn't fail if there aren't any tests

Piedone:
[MENT-157](https://lombiq.atlassian.net/browse/MENT-157)


.github/actions/test-dotnet/action.yml
```diff
@@ -75,13 +76,13 @@ runs:
     # The same is true for the next step as well.
     - name: Merge FailureDumps
       shell: pwsh
-      if: success() || failure()
+      if: (success() || failure()) && steps.run-tests.outputs.test-count != 0
```
BenedekFarkas, line 85: I wanted to ask if this can be moved to a reusable (for later steps) value, but I guess that won't work, because the value comes from a previous step's output?

Piedone: `success()` and `failure()` needs to be evaluated for each step, yes.

BenedekFarkas: I meant the `steps.run-tests.outputs.test-count != 0` part.

Piedone: That could at most be simplified to an env var (i.e. fewer dots on the expression) but I don't see much of a point in that since the boolean expression needs to remain (GHA YAML syntax needs it).

PR #129, opened 2022-11-29
HAST-290: New build-and-test-dotnet workflow that also runs tests in addition to a .NET build

Piedone:
[HAST-290](https://lombiq.atlassian.net/browse/HAST-290)


.github/actions/test-dotnet/action.yml
```diff
@@ -66,6 +73,7 @@ runs:
       working-directory: ${{ inputs.build-directory }}
       run: |
         $switches = @{
+          Solution = (Get-ChildItem ${{ inputs.solution-path }}).FullName
```
sarahelsaig: If you add ` | Select-Object -First 1` inside the parentheses then it will never break. Wouldn't that be preferable?

sarahelsaig (changes requested):

.github/actions/test-dotnet/action.yml
Piedone: Do you mean instead of having a `solution-path` parameter? While that would make it possible to run this action regardless of there being multiple solutions, it wouldn't solve us wanting to run tests only for a selected one.

sarahelsaig: Why, I didn't say that. My concern is that right now if you forget the `solution-path` then you'll get an error message from `dotnet` which doesn't offer any help with the correct workflow invocation so it will be confusing for GHA users who aren't already familiar with this input.

An alternative would be, to extract this into a variable before `$swiches`, check its count, and if it's not 1 then write an error message with the current item count and list the items out. (You'd need the count too, because if you set `solution-path` but it's wrong, e.g. typo, then the 0 count is very relevant info.) Then `exit 1`.

Piedone: I see, so you're talking about scenarios when there are multiple solutions but `solution-path` is not set.

PR #136, opened 2022-12-08
OSOE-490: Create binlog file during build and add it to build artifacts

### LINKED ISSUE #132: Create binlog file during build and add it to build artifacts (OSOE-490)
This allows to inspect the inner workings of any build, no matter the configured `verbosity`.

Needs this additional parameter on `dotnet build`:

```pswh
-bl:build.binlog
```

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-490)

Piedone:
[OSOE-490](https://lombiq.atlassian.net/browse/OSOE-490)
Fixes #132.

### DIFF
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -152,5 +152,5 @@ runs:
       if: (success() || failure()) && inputs.create-binary-log == 'true'
       with:
         name: build.binlog
-        path: build.binlog
+        path: ${{ inputs.directory }}/build.binlog
         if-no-files-found: ignore
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -154,6 +154,6 @@ runs:
       uses: actions/upload-artifact@v3.1.1
       if: (success() || failure()) && inputs.create-binary-log == 'true'
       with:
-        name: build-binary-log-${{ steps.build.outputs.artifact-name-suffix }}.binlog
+        name: build-binary-log-${{ steps.build.outputs.artifact-name-suffix }}
         path: ${{ inputs.directory }}/build.binlog
         if-no-files-found: ignore
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -60,6 +60,12 @@ inputs:
       If not empty, it indicates that the static code analysis should fail. Then each line should contain one error
       code. The static code analysis should yield these and only these errors. This is ignored if `enable-code-analysis`
       is not "true". You may add a colon and then an explanation of the expected error.
+  create-binary-log:
+    required: false
+    default: "false"
+    description: >
+      Configures MSBuild to create a binary log. This is useful to inspect and debug builds. See
+      https://github.com/dotnet/msbuild/blob/main/documentation/wiki/Binary-Log.md for details.
 
 runs:
   using: "composite"
@@ -132,6 +138,7 @@ runs:
             Version = "${{ inputs.publish-version }}"
             Switches = $buildSwitches
             ExpectedCodeAnalysisErrors = $expectedCodeAnalysisErrors
+            CreateBinaryLog = $${{ inputs.create-binary-log }}
         }
 
         $startTime = [DateTime]::Now
@@ -139,3 +146,11 @@ runs:
         $endTime = [DateTime]::Now
 
         Write-Output ("Solution or project build took {0:0.###} seconds." -f ($endTime - $startTime).TotalSeconds)
+
+    - name: Upload MSBuild Binary Log
+      uses: actions/upload-artifact@v3.1.1
+      if: (success() || failure()) && inputs.create-binary-log == 'true'
+      with:
+        name: build.binlog
+        path: build.binlog
+        if-no-files-found: ignore
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -124,7 +124,7 @@ runs:
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
-        Initialize-ArtifactNameSuffix '${{ inputs.build-directory }}'
+        Initialize-ArtifactNameSuffix '${{ inputs.directory }}'
 
         $buildSwitches = @'
             ${{ inputs.dotnet-build-switches }}
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -154,6 +154,6 @@ runs:
       uses: actions/upload-artifact@v3.1.1
       if: (success() || failure()) && inputs.create-binary-log == 'true'
       with:
-        name: build-binary-log-${{ steps.build.outputs.artifact-name-suffix }}
+        name: build-binary-log-${{ steps.build.outputs.artifact-name-suffix }}.binlog
         path: ${{ inputs.directory }}/build.binlog
         if-no-files-found: ignore
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -120,9 +120,12 @@ runs:
         restore-keys: ${{ steps.detect-cache-configuration.outputs.restore-keys }}
 
     - name: Build and Static Code Analysis
+      id: build
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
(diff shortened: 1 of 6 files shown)


0liver (approved):

.github/actions/build-dotnet/action.yml
```diff
@@ -132,10 +141,19 @@ runs:
...
         Build-DotNetSolutionOrProject @switches
         $endTime = [DateTime]::Now
 
         Write-Output ("Solution or project build took {0:0.###} seconds." -f ($endTime - $startTime).TotalSeconds)
+
+    - name: Upload MSBuild Binary Log
+      uses: actions/upload-artifact@v3.1.1
+      if: (success() || failure()) && inputs.create-binary-log == 'true'
+      with:
+        name: build-binary-log-${{ steps.build.outputs.artifact-name-suffix }}
```
0liver: Where do these suffixes (1 and 2) come from? Cannot find them in the code. Also, why are they so different?

image

1. Seems superfluous.
2. Contains spaces - not ideal. And also superfluous.

0liver (changes requested):

.github/actions/build-dotnet/action.yml
0liver: Also, while the `.binlog` suffix might not be necessary, it helps identify the file type when scanning the assets. Not having it somehow looks unintuitive:

image

.github/actions/build-dotnet/action.yml
```diff
@@ -132,10 +141,19 @@ runs:
...
         }
 
         $startTime = [DateTime]::Now
         Build-DotNetSolutionOrProject @switches
         $endTime = [DateTime]::Now
 
         Write-Output ("Solution or project build took {0:0.###} seconds." -f ($endTime - $startTime).TotalSeconds)
+
+    - name: Upload MSBuild Binary Log
+      uses: actions/upload-artifact@v3.1.1
```
0liver, line 154: Is there a way to disable zipping the binlog? Because it doesn't save any space, and is just cumbersome when downloaded. (I'm exaggerating a bit, of course, but it would still be nice to simply download the file with the name instead of a zip archive.)

0liver (changes requested):

.github/actions/build-dotnet/action.yml
Piedone: Artifacts need to have unique names for the whole workflow. Not adding OS-specific and run-specific suffixes will make them overwrite each other. Since we have two pairs of runs in the same workflow, we need to be able to distinguish between them.

The names come from `Initialize-ArtifactNameSuffix`

Are you scanning from the end of lines? :) The artifact names start with what the artifact is about ("build-binary-log" and "ui-test-failure-dump") so I don't see why we'd want to also add the same thing at the end, creating _build-binary-log-<suffix>.binlog.zip_ files.

.github/actions/build-dotnet/action.yml
Piedone: No, having a zip is mandatory: https://github.com/actions/upload-artifact#zipped-artifact-downloads

.github/actions/build-dotnet/action.yml
0liver: I highlighted the parts of the artifact names that were unclear to me.

I saw your code about artifact suffixes, and they make a lot of sense. But I'm wondering about the hex hash at the end of the first two artifact names and the "GitHub Actions N" at the end of the other two. Where do **those** come from, and why are they of different types?

Regarding the `.binlog` suffix - yes, I scan file extensions maybe even faster than the file name. Usually, the file extensions tells me all I need to know, e.g. in email attachments I usually want to open the PDF not the s7mime file. I'm really in favor of having the suffix there - seeing `.binlog` at the end I know exactly that this is an MSBuild binary log file, no thinking involved. Reading "build-binary-log-something" is, well, just a name that I can give to anything. It could be a screenshot, as well.

.github/actions/build-dotnet/action.yml
0liver: Thank you. Simple API, good API.

.github/actions/build-dotnet/action.yml
Piedone: These names all come from here: https://github.com/Lombiq/GitHub-Actions/blob/e0a7bae4129d6ab63e69fed924d60679e0b56e8f/Scripts/Initialize-ArtifactNameSuffix.ps1#L13 Since the hash-like part is the last one, it's [the `RUNNER_NAME` context of GitHub](https://docs.github.com/en/actions/learn-github-actions/environment-variables#:~:text=own%20job%20steps.-,RUNNER_NAME,-The%20name%20of).

I don't know of any better way to name artifacts that meet all of these criteria:

- Unique for a given workflow run, even if multiple jobs run under the same OS with just matrix parameters changing.
- Has some information that'll let you correlate it with the specific job and matrix execution (one like "Build and Test - NuGetTest solution (ubuntu-latest, 20)" - note that a name like this, specifically, is not available anywhere in the context nor the information to construct it, though that would be ideal).
- Does this in a human-readable way, meaning that you'll be able to do this at a glance, without having to look up IDs or something. E.g. while "build-binary-log-NuGetTest-Linux-X64-GitHub Actions 6" is ugly, taking a look at the executed jobs, you'll quickly realize what it relates to; barring very curious cases, you'll be able to just eyeball it (sometimes you'll need to open the job details).

Check out the dump of the whole GitHub context under each job here: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3651821161/jobs/6172132873 I don't see any better information here that would help us do this. I you do, please provide a suggestion.

Why would you name anything "build-binary-log" that is not specifically a binary log for a build? :) Technically, of course, it needn't be, but also technically, a ".binlog" suffix at the end is just part of the name as well (and not an extension). But adding a suffix to these files' names is really not something I think we should take any more time debating: I've added it.

0liver: 1. Thanks for adding the suffix. I do think that it adds value and we should do it generally.
2. Thanks for the GITHUB_CONTEXT output.
3. Trying to cut down the noise, I think `$runnerSuffix` could simply become `"$Env:RUNNER_NAME".Replace(' ', '-')`. Having `$Env:RUNNER_OS` and `$Env:RUNNER_ARCH` in there don't seem to add any value, since the OS is already part of `RUNNER_NAME`, and the architecture is not really needed here.
4. For the artifact name, having the `.binlog` suffix in place, I think we can remove `binary-log` altogether and shorten it to:
   ```gha
       name: build-${{ steps.build.outputs.artifact-name-suffix }}.binlog
   ```

0liver (changes requested):

.github/actions/build-dotnet/action.yml
Piedone: 3. `RUNNER_NAME` is not enough. Most of the time it's indeed something like "windows-latest-8core_4c10d9439cf2". However, frequently, it's something like "GitHub Actions 2" that won't help you, like [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3644369794) (this is seemingly random, can happen for Linux and Windows builds, as well as root solution or NuGetTest). `RUNNER_OS` is thus necessary to distinguish between two builds in the same directory. That, in turn, is not enough alone, since you can have multiple different images all running on Windows, or a different version of Windows, hence we need it in combination with `RUNNER_NAME`. Finally, `RUNNER_ARCH` is also necessary, because there can be 32b runners, and between e.g. a 32b and 64b Windows runner with "GitHub Actions 2"-like `RUNNER_NAME`s we otherwise wouldn't be able to identify.

While sure, we could add dashes to the name, I don't really see why (or are we really talking about aesthetics here?), and that would only increase the chance of a name collision, removing entropy (how the agent name is generated we don't know, and just as we have dashed names and spaced names, there can be a combination of these at one time, at which point we'll need to debug why an artifact is not showing up).

I really think there is nothing to improve here.

4. Why would we remove information from the name? These are things that should be clear to anyone. An extension-like non-extension doesn't help that. While I personally see no point in adding that, as I said before, it helps you, probably others too, so let's add it. But I see value removed by using that as the only means to communicate what this zip contains.

0liver: Thank you for taking the time to explain that. I am convinced now. Let's merge this.

0liver (approved):

.github/actions/build-dotnet/action.yml
Piedone: I realize it's not an obvious topic, no problem.

PR #139, opened 2022-12-08
Revert "OSOE-470: Remove false entries (actual typos) from spell checking allow lists in Lombiq.GitHub.Actions"

BenedekFarkas:
Reverts Lombiq/GitHub-Actions#111


Piedone:
Please be sure to only merge submodules to dev just right before you merge the superproject PR too.

BenedekFarkas:
I would've if the superproject PR was working correctly with this merge, which it didn't, that's why this merge was reverted.

Piedone:
OK!

PR #140, opened 2022-12-08
NEST-325: Add Application Insights release annotations to deployments

Piedone:
[NEST-325](https://lombiq.atlassian.net/browse/NEST-325)


.github/actions/add-azure-application-insights-release-annotation/action.yml
```diff
@@ -0,0 +1,36 @@
...
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+
+    - name: Add Azure Application Insights Release Annotation
+      shell: pwsh
+      env:
+        # Might contain user input so should go via an env var for security.
+        RELEASE_NAME: ${{ inputs.release-name }}
+      run: |
+        $params = @{
+            ApplicationInsightsResourceId = '${{ inputs.application-insights-resource-id }}'
```
wAsnk, line 32: Shouldn't we use the same indentation rule (2 spaces) here as well, so we use the same all over the yml files? Or did you deliberately do this because it's a powershell script? Same question in _add-jira-issue-code-to-pull-request/action.yml_

Piedone: Yes, this is deliberate due to being PS code. 2-spaces is only needed (and actually required syntactically) for YML.

wAsnk (approved):

PR #141, opened 2022-12-09
OSOE-470: Remove false entries (actual typos) from spell checking allow lists

BenedekFarkas:
[OSOE-470](https://lombiq.atlassian.net/browse/OSOE-470)


.github/actions/spelling/patterns.txt
```diff
@@ -1,21 +1,21 @@
...
+sha[0-9]{3}-[0-9a-zA-Z+/]{64}
+
+# Orchard Core content item ID.
+\b[0-9a-z]{26}\b
+
+# Git commit hashes.
+\b[0-9a-f]{40}\b
+
+# Hex color codes.
+#[0-9a-fA-F]{3,6}
```
BenedekFarkas: @0liver do you have an idea why this doesn't ignore `AFAFAF`, but it works with all the other ones in https://github.com/Lombiq/Orchard-UIKit/blob/dev/Lombiq.UIKit/Assets/Styles/textbox.scss#L15?

See: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3659589978/jobs/6185754308#step:5:103

We can ignore that line, but I'd prefer properly fixing this, but I'm out of ideas. Thanks!

0liver: Tbh, I don't know, especially after Zoltán mentioned that the patterns are supposed to match **lines** that will be ignored. You could try to switch the two values in _textbox.css_ and see whether that makes a difference.

BenedekFarkas: Actually, it turns out they don't, see the 2nd point in: https://lombiq.atlassian.net/browse/OSOE-470?focusedCommentId=72613
Thanks for tip, I'll try that!

PR #143, opened 2022-12-12
OSOE-423: Add new spelling exceptions

Luko6:
[OSOE-423](https://lombiq.atlassian.net/browse/OSOE-423)

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -77,6 +77,7 @@ dragleave
 dragover
 endcapture
 extname
+fileupload
 flyout
 formattable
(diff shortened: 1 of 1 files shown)


.github/actions/spelling/allow.txt
```diff
@@ -77,6 +77,7 @@ dragleave
 dragover
 endcapture
 extname
+fileupload
```
Piedone, line 80: Instead of adding these rare exceptions, use the new `#spell-check-ignore-line` pattern at the end of the affected lines to exclude them inline.

Piedone (changes requested):

PR #153, opened 2022-12-20
NEST-337: Re-implement Azure Swap build with GitHub Actions

DemeSzabolcs:
[NEST-337](https://lombiq.atlassian.net/browse/NEST-337)


.github/workflows/swap-azure-web-app-slots.yml
```diff
@@ -0,0 +1,103 @@
...
+    inputs:
+      cancel-workflow-on-failure:
+        description: When set to "true", will cancel the current workflow run with all jobs if this workflow fails.
+        required: false
+        type: string
+        default: "true"
+      machine-type:
+        required: false
+        type: string
+        default: ubuntu-latest
```
Piedone: Use `ubuntu-22.04` as the other workflows.

.github/workflows/swap-azure-web-app-slots.yml
```diff
@@ -0,0 +1,103 @@
...
+      run:
+        shell: pwsh
+    timeout-minutes: ${{ inputs.timeout-minutes }}
+    steps:
+    - name: Login to Azure
+      # v1.4.6
+      uses: azure/login@92a5484dfaf04ca78a94597f4f19fea633851fa2
+      with:
+        creds: ${{ secrets.AZURE_APP_SERVICE_SWAP_SERVICE_PRINCIPAL }}
+        enable-AzPSSession: true
```
Piedone: We use this in multiple places now. Refactor it to a common internal action, like `checkout`.

.github/workflows/swap-azure-web-app-slots.yml
```diff
@@ -0,0 +1,103 @@
...
+
+    - name: Initialize PowerShell modules
+      uses: Lombiq/Infrastructure-Scripts/.github/actions/initialize@dev
+
+    - name: Add Azure Application Insights Release Annotation
+      uses: Lombiq/GitHub-Actions/.github/actions/add-azure-application-insights-release-annotation@dev
+      with:
+        application-insights-resource-id: ${{ inputs.application-insights-resource-id }}
```
Piedone: Only run this after the swap so it's only added if the swap actually succeeds.

Piedone (changes requested):

.github/workflows/swap-azure-web-app-slots.yml
DemeSzabolcs: We already have an action called `checkout`. Do you mean, that I should add it to that action as an optional step?

Piedone: Encapsulate this code here into a composite action, similar to how we did with our `checkout` action, and use it instead of `azure/login` everywhere.

DemeSzabolcs: Oh okay, I misunderstood.

.github/workflows/swap-azure-web-app-slots.yml
```diff
@@ -96,6 +89,11 @@ jobs:
           -WebAppName ${{ inputs.app-name }} `
           -SlotName ${{ inputs.destination-slot-name }}
 
+    - name: Add Azure Application Insights Release Annotation
+      uses: Lombiq/GitHub-Actions/.github/actions/add-azure-application-insights-release-annotation@dev
+      with:
+        application-insights-resource-id: ${{ inputs.application-insights-resource-id }}
+
```
Piedone: This should rather be run after the swap itself. If the ping fails, then that's bad but the app is still out in the slot, so a release annotation is due.

Docs/Actions.md
```diff
@@ -14,6 +14,7 @@ In addition to the below short explanations, check out the inline documentation
 - `check-merge-conflict`: Labels and comments on pull requests with merge conflicts.
 - `create-jira-issues-for-community-activities`: Creates Jira issues for community activities happening on GitHub, like issues, discussions, and pull requests being opened. Pull requests are only taken into account if they're not already related to a Jira issue (by starting their title with a Jira issue key).
 - `enable-corepack`: Enables [Node corepack](https://nodejs.org/docs/latest-v16.x/api/corepack.html) so any package manager can be used seamlessly.
+- `login-to-azure`: Logs in to Azure using a service principal and the [azure/login action](https://github.com/Azure/login).
```
Piedone: Similar to `checkout`, this should be an internally-only used action (there's no point in using it for others), without being advertised here. Also, similarly, add a not to the action itself.

.github/actions/login-to-azure/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Login to Azure
+description: Logs in to Azure using a service principal.
+
+inputs:
+  enable-azure-powershell-session:
+    required: false
+    default: "true"
+    description: Whether to enable an Azure PowerShell session.
+  service-principal:
```
Piedone: This is a secret and should be passed around as such, in case of composite actions as an env var. See e.g. how it works with `JIRA_API_TOKEN` from the `create-jira-issues-for-community-activities` workflow.

.github/actions/login-to-azure/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Login to Azure
+description: Logs in to Azure using a service principal.
+
+inputs:
+  enable-azure-powershell-session:
+    required: false
+    default: "true"
+    description: Whether to enable an Azure PowerShell session.
```
Piedone:
```suggestion
    description: Whether to enable an Az PowerShell session.
```

.github/actions/login-to-azure/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Login to Azure
+description: Logs in to Azure using a service principal.
+
+inputs:
+  enable-azure-powershell-session:
```
Piedone: This is an "Az" PS session. It's important, because that's not the same as the "Azure PowerShell module".

.github/actions/login-to-azure/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Login to Azure
+description: Logs in to Azure using a service principal.
```
Piedone: Why would you use this, what kind of login? What happens after a login?

Piedone (changes requested):

Piedone (approved):

PR #161, opened 2022-12-23
OSOE-508: Every workflow that uses our checkout action should accept (and pass on) the CHECKOUT_TOKEN secret

### LINKED ISSUE #160: Every workflow that uses our checkout action should accept (and pass on) the CHECKOUT_TOKEN secret (OSOE-508)
To be able to clone authenticated submodules.

For example, `build-and-test-dotnet` and `build-and-test-orchard-core` already do, but the following don't:

1. build-dotnet (although it's deprecated, worth updating, just in case)
2. deploy-to-azure-app-service
3. publish-nuget
4. test-analysis-failure

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-508)

Related: https://github.com/actions/runner/issues/924

BenedekFarkas:
[OSOE-508](https://lombiq.atlassian.net/browse/OSOE-508)
Fixes #160

### DIFF
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -3,6 +3,13 @@ name: Build .NET solution
 
 on:
   workflow_call:
+    secrets:
+      CHECKOUT_TOKEN:
+        required: false
+        description: >
+          The GitHub token to authenticate checkout. Pass in a GitHub personal access token if authenticated submodules
+          are used.
+
     inputs:
       cancel-workflow-on-failure:
         description: When set to "true", will cancel the current workflow run with all jobs if this workflow fails.
@@ -96,6 +103,8 @@ jobs:
 
     - name: Checkout
       uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+      with:
+        token: ${{ env.CHECKOUT_TOKEN }}
 
     - name: Set up .NET
       uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -104,7 +104,7 @@ jobs:
     - name: Checkout
       uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
       with:
-        token: ${{ env.CHECKOUT_TOKEN }}
+        token: ${{ secrets.CHECKOUT_TOKEN }}
 
     - name: Set up .NET
       uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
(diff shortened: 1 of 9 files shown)


.github/workflows/build-dotnet.yml
```diff
@@ -96,6 +103,8 @@ jobs:
 
     - name: Checkout
       uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+      with:
+        token: ${{ env.CHECKOUT_TOKEN }}
```
Piedone: This will be empty by default, and thus no authentication token will be passed to `actions/checkout` (overwriting the default `github.token`). This will fail checkout if the current repository is authenticated. Do what e.g. `build-and-test-orchard-core` does, which allows using `github.token` if no other token is specified (which is the desired case if there are no private submodules).

Perhaps that logic can be somewhat DRYd, but I'm not sure how much, since composite actions need secrets passed in explicitly, they can't access them from the context.

Unless `actions/checkout` takes an empty string parameter as no parameter and will use `github.token` instead; then the default-handling logic in other workflows is unnecessary, so please remove that.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -3,10 +3,16 @@ name: Deploy to Azure App Service
 on:
   workflow_call:
     secrets:
+      CHECKOUT_TOKEN:
```
Piedone, line 6: Please open a branch/PR in NEST so we can test this.

Piedone (changes requested):

.github/workflows/build-dotnet.yml
BenedekFarkas: Gotcha, I'll test in NEST then like this and we'll see which way to change it!

BenedekFarkas: We'll probably need to use that logic, because 3 out of 4 other workflows that already use the secret has it. The 4th one is test-analysis-failure, but it's not used in a situation yet where the token is needed, that's why it didn't come up there. I'll update that too.

BTW the logic can't be factored out to our checkout action?

Piedone: I'm not entirely sure if the default-handling logic is actually needed, so please do check.

You can actually check for an empty `token` input in our `checkout` action and use `github.token` (or `$Env:GITHUB_TOKEN`) if it's empty, because I think the GitHub token, unlike any secret, is actually available in composite actions too. And then you can just directly pass `CHECKOUT_TOKEN` to `token` in all consumers without a separate "Set Checkout Token" step.

Piedone: No, `$Env:GITHUB_TOKEN` is not available in actions by default, needs to be explicitly passed in, see e.g. `post-pull-request-checks-automation`.

BenedekFarkas: Yeah, my thoughts exactly - copy-paste the same step and just shuffle around some of the variables.
It seems to work, I just made a mistake in the deployment workflow at first, otherwise it will be good to go and all the workflows calling the checkout action can be simplified.

Piedone (approved):

PR #170, opened 2022-12-28
OSOE-518: Fix formatting of action/workflow files

### LINKED ISSUE #169: Fix formatting of action/workflow files (OSOE-518)
And include recommendation for editing files (VS Code + https://marketplace.visualstudio.com/items?itemName=me-dutour-mathieu.vscode-github-actions) in the documentation.

Static analysis would also be nice, although this plugin still has false positives, because the underlying language server is not GHA-specific, but the GH team is working on an official one: https://github.com/github/roadmap/issues/564

Maybe https://github.com/oxsecurity/megalinter/blob/main/docs/descriptors/yaml.md?

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-518)

BenedekFarkas:
[OSOE-518](https://lombiq.atlassian.net/browse/OSOE-518)
Fixes #169

The types of changes in this PR:

1. Removing extra whitespaces (end of line space, double spaces).
2. Fixing indentation (usually adding one more for list items and their descendants).
3. Apostrophe vs. double quotes (mostly with autoformatting, but manually in descriptions).
4. Line wrapping in descriptions.

Reset Azure Environment includes each type of change and has the most changes overall, here's the test run in NEST: https://github.com/Lombiq/DotNest-Tenants-Core/actions/runs/3795923016/jobs/6455518740

### DIFF
--- a/.github/workflows/deploy-to-azure-app-service.yml
+++ b/.github/workflows/deploy-to-azure-app-service.yml
@@ -80,7 +80,7 @@ on:
       slot-name:
         required: true
         type: string
-        description:
+        description: >
           What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
           name in parenthesis.
       resource-group-name:
@@ -104,7 +104,7 @@ on:
         required: false
         type: string
         default: "false"
-        description:
+        description: >
           Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
           target machine.
       application-insights-resource-id:
--- a/.github/workflows/deploy-to-azure-app-service.yml
+++ b/.github/workflows/deploy-to-azure-app-service.yml
@@ -80,7 +80,8 @@ on:
       slot-name:
         required: true
         type: string
-        description: What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
+        description:
+          What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
           name in parenthesis.
       resource-group-name:
         required: true
@@ -103,7 +104,8 @@ on:
         required: false
         type: string
         default: "false"
-        description: Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
+        description:
+          Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
           target machine.
       application-insights-resource-id:
         required: true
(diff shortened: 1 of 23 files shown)


.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -80,7 +80,8 @@ on:
       slot-name:
         required: true
         type: string
-        description: What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
+        description:
+          What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
```
Piedone: While this doesn't matter much in practice right now, this adds a line break to the value of `description`, as opposed to the value being a single-line string (what we'd need if this were displayed in some UI) and only having line for the editing experience..

```suggestion
        description: >
          What you see at the top of the blade on the Azure Portal, when you open the slot, before the app
```

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -103,7 +104,8 @@ on:
         required: false
         type: string
         default: "false"
-        description: Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
+        description:
+          Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
```
Piedone:
```suggestion
        description: >
          Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the
```

Readme.md
```diff
@@ -13,4 +15,6 @@ Some common workflows and actions for Github Actions shared between Lombiq proje
 
 Bug reports, feature requests, comments, questions, code contributions and love letters are warmly welcome. You can send them to us via GitHub issues and pull requests. Please adhere to our [open-source guidelines](https://lombiq.com/open-source-guidelines) while doing so.
 
+We recommend using [Visual Studio Code](https://code.visualstudio.com/) with the (unofficial) [GitHub Actions](https://marketplace.visualstudio.com/items?itemName=me-dutour-mathieu.vscode-github-actions) extension (until [the official one](https://github.com/github/roadmap/issues/564) is released).
```
Piedone: This is better suited for https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions (we don't just write GHA code in this repo). Also include any notable rules for things you made uniform.

Piedone (changes requested):

.github/workflows/deploy-to-azure-app-service.yml
BenedekFarkas: I agree! I was experimenting with some settings and modified this one with line wrapping set to "preserve", which caused it not to move the beginning of the text to first line even without `>`.
Auto-wrapping can be improved even further by changing a setting in VS Code, I'll try that and add it to the instructions.

Readme.md
BenedekFarkas: Yep, I didn't forget that, but wasn't sure yet how to divide up the information. I'd actually expand this information here (there's not really much to add in terms of how to write code, just configuring the plugin and auto-formatting will take care of it) based on what I just changed locally and link this section of the readme from there (and any other source), so external contributors can access it too.

BenedekFarkas: Update: There are some side-effects to what I wanted to set up in the GHA extension:

- When you set `Prose Wrap` to `always` and `Print Width` to `120`, it will correctly wrap lines the way we want, but auto-formatting in the unofficial GHA extension is not entirely context-aware and it (figuratively and literally) breaks code lines too, not just descriptions, so that's a no-go.
- `Prose Wrap` set to `preserve` works if you go over the line width but doesn't rewrap the text if a line still has space for text from the next line.
- But the Rewrap extensions still handles this correctly, because it's just text!

So, my suggestions:

1. Keep using the unofficial GHA extension, because it's great for schema validation and IntelliSense, but leave the wrapping settings on default, because we don't actually use them.
2. Keep using Rewrap (that's already in our VS Code guidelines) for wrapping descriptions.
3. The previous also makes it necessary to start one-line descriptions on a new line (and put ` <` after `description:`), because Rewrap isn't context-aware either, so if the description grows over the length limit (e.g., you start with `description: Hello Lombiq!` and expand it), then wrapping will kick in and also move `description:` to the previous line.
This is not a problem when the description is one line but starts with `<` and the text is on the next line or when the text is already multi-line (regardless of having `<` or not).
4. To generalize the previous point, I'd rather just say that description should always have the `<` for consistency and start the text on the next line.

I'd add these here in the readme (in a more concise way) and we can link it where it's relevant.

Piedone: Adding a character to the code for step 3 that otherwise isn't needed just to make an IDE plugin work correctly is not something I'd do. Since wrapping is a trivial thing anyway (and works with Rewrap as you also point out), and it's not needed for single-line descriptions (if you move to multi-lines then wrapping works anyway) I don't see why we'd set this up.

BenedekFarkas: My though process was more about consistency, i.e., to enforce a styling that always looks the same (not just for the sake of the extension not tripping up but regarding YAML-syntax in general too). The problem is more about not being able to enforce it and GHA doesn't care how we fold descriptions (they won't be rendered anywhere), so the best we can do reliably is just a few instructions (mostly about Rewrap) - stay tuned.

Piedone: We can perhaps have proper enforcement with https://github.com/Lombiq/NodeJs-Extensions/issues/11.

BenedekFarkas: Yep, Oliver commented this on #169, that would be a great addition!

Readme.md
```diff
@@ -14,3 +16,10 @@ Some common workflows and actions for Github Actions shared between Lombiq proje
 Bug reports, feature requests, comments, questions, code contributions and love letters are warmly welcome. You can send them to us via GitHub issues and pull requests. Please adhere to our [open-source guidelines](https://lombiq.com/open-source-guidelines) while doing so.
 
 This project is developed by [Lombiq Technologies](https://lombiq.com/). Commercial-grade support is available through Lombiq.
+
+### Code styling
```
Piedone: I still think this is better in our wiki, especially that we actually already have a section for this: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=LGHA)%20for%20examples.-,Editor%20support,-%3A These notes are not specific to this project, and on the other hand, the same practices need to be followed by other projects too.

Piedone (changes requested):

Readme.md
BenedekFarkas: I don't disagree with that especially for keeping the same type of information in the same place, but that still means that this information is not visible outside Lombiq. Which might not be an issue now, because we don't have external contributors at the moment (but hopefully we will). On the other hand, YAML linting might alleviate that (but it would still be better, if the code styling rules were clear upfront, not just from e.g., a failed build, but we can worry about that later).

PR #173, opened 2023-01-04
NEST-392: Fix and improve spell-checking configuration in DotNest Core SDK

BenedekFarkas:
[NEST-392](https://lombiq.atlassian.net/browse/NEST-392)

[NEST-392]: https://lombiq.atlassian.net/browse/NEST-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NEST-392]: https://lombiq.atlassian.net/browse/NEST-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/spelling/excludes.txt
+++ b/.github/actions/spelling/excludes.txt
@@ -7,10 +7,10 @@
 \.editorconfig$
 \.eslintrc$
 \.git
-\.htmlvalidate\.json$
 \.ico$
 \.jfif$
 \.jpg$
+\.json$
 \.min\.css$
 \.min\.js$
 \.ndproj$
@@ -21,7 +21,6 @@
 \.png$
 \.ps1$
 \.psm1$
-\.recipe\.json$
 \.ruleset$
 \.sln$
 \.stylelintrc$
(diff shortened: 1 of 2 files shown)


.github/actions/spelling/excludes.txt
```diff
@@ -7,10 +7,10 @@
 \.editorconfig$
 \.eslintrc$
 \.git
-\.htmlvalidate\.json$
 \.ico$
 \.jfif$
 \.jpg$
+\.json$
```
Piedone, line 13: JSON files can have comments, and it would be useful to spell-check only those. Isn't that possible?

Piedone (changes requested):

.github/actions/spelling/excludes.txt
BenedekFarkas: I was wondering the same thing, but AFAIK "ignore everything in .json files except comments (based on regex match)" is not possible.
I'm not married to the idea of ignoring every JSON, comments are a good counterargument, but it looks like an OK compromise with all the various files we were explicitly listing.

PR #178, opened 2023-01-10
OSOE-531: Experimental spell-checking ignore pattern for interfaces and generic type parameters

### LINKED ISSUE #177: Experimental spell-checking ignore pattern for interfaces and generic type parameters (OSOE-531)
allow.txt has a bunch of entries like `IExample` and `TExample`. Let's investigate a pattern that ideally removes the I and T letters from the beginning of these, so that the rest can be spell-checked normally.

Pro: Getting rid of almost 200 entries from allow.txt.
Con (kind of): Mistyping `Illinois` to `ILlinois` or `Temperature` to `TEmperature` also matches the regex. However, the remaining words would be `Llinois` and `Emperature`, so you'd end up with a typo you need to check anyway.

In any case, I'll add instructions to advice.md that whatever comes up in the spell-checking report, you should always check out the detailed log to see where the actual entry is coming from and what original text is.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-531)

BenedekFarkas:
[OSOE-531](https://lombiq.atlassian.net/browse/OSOE-531)
Fixes #177

Regex test cases - first block doesn't match (as expected), second block matches (as expected), except for `IICantHandleThis`, but that naming scheme is an edge case (and we currently don't have any such cases detected in any of the consumer projects):

```
information
template
superproject
JObject
JObjectGenerator
Illinois
Temperature
ILLINOIS
TEMPERATURE
IEDriver
IIEDriver
TIEDriver
IOException
IIOException
TIOException

ILlinois
TEmperature
IVue
TClass
IVueWaitWeb
TWorkflowAttribute
ICantBelieveItsNotDataTable
IICantHandleThis
```

[OSOE-531]: https://lombiq.atlassian.net/browse/OSOE-531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/spelling/patterns.txt
```diff
@@ -20,5 +20,11 @@ sha[0-9]{3}-[0-9a-zA-Z+/]{64}
 # Hexadecimal literals in C# code.
 \b0x[0-9a-fA-F]+\b
 
+# Match the "I" from the beginning of interface names and the "T" from the beginning of generic type parameter names.
+# The pattern matches, if the first letter is "I" or "T", followed a capital letter, followed by a word character that
```
0liver:
```suggestion
# The pattern matches, if the first letter is "I" or "T", followed by a capital letter, followed by a word character that
```

.github/actions/spelling/patterns.txt
```diff
@@ -20,5 +20,11 @@ sha[0-9]{3}-[0-9a-zA-Z+/]{64}
...
+# is not a capital letter, followed by 0 or more word characters. Only the first letter is captured, so the rest is
+# still spell-checked as they are usually made up of actual words.
+\b(I|T)(?=[A-Z][a-z0-9_][\w]*)
```
0liver: Why would you allow digits after the capital letter?

IA7bla
TB3foo

?

0liver: For our purposes, something stricter would probably be simpler since this is about interface and class names:
```suggestion
\b(I|T)(?=[A-Z][a-z]\w*)
```
I'd keep the `\w` (no need for parentheses) for simplicity, though, too 😉

BenedekFarkas: .NET/C# conventions allow numbers and some special characters in interface/class names. It's rare for us but can happen with an external library that we reference in our code. I'll remove the brackets around `\w`!

porgabi (approved):

PR #179, opened 2023-01-10
OSOE-501: Enable further analyzer rules for better code styling control and more in Lombiq.Analyzers.PowerShell

BenedekFarkas:
[OSOE-501](https://lombiq.atlassian.net/browse/OSOE-501)

[OSOE-501]: https://lombiq.atlassian.net/browse/OSOE-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/add-azure-application-insights-release-annotation/Add-ReleaseAnnotation.ps1
```diff
@@ -7,11 +7,11 @@ param(
 Write-Output "Adding release annotation with the release name `"$ReleaseName`"."
 
 $annotation = @{
-    Id = [GUID]::NewGuid();
-    AnnotationName = $ReleaseName;
-    EventTime = (Get-Date).ToUniversalTime().GetDateTimeFormats("s")[0];
+    Id = [GUID]::NewGuid()
```
sarahelsaig: Weird all-caps type name...

```suggestion
    Id = [Guid]::NewGuid()
```

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -10,16 +10,16 @@ param ($Solution, $Verbosity, $Filter, $Configuration)
...
 # default buffer width.
 
-$ConnectionStringSuffix = ";MultipleActiveResultSets=True;Connection Timeout=60;ConnectRetryCount=15;ConnectRetryInterval=5;TrustServerCertificate=true;Encrypt=false";
+$ConnectionStringSuffix = ";MultipleActiveResultSets=True;Connection Timeout=60;ConnectRetryCount=15;ConnectRetryInterval=5;TrustServerCertificate=true;Encrypt=false"
 if ($Env:RUNNER_OS -eq "Windows")
 {
-    $Env:Lombiq_Tests_UI__SqlServerDatabaseConfiguration__ConnectionStringTemplate =
-        "Server=.\SQLEXPRESS;Database=LombiqUITestingToolbox_{{id}};Integrated Security=True" + $ConnectionStringSuffix
+    $Env:Lombiq_Tests_UI__SqlServerDatabaseConfiguration__ConnectionStringTemplate = ("Server=.\SQLEXPRESS;" +
+        "Database=LombiqUITestingToolbox_{{id}};Integrated Security=True" + $ConnectionStringSuffix)
```
sarahelsaig: Why is this necessary? imo it looks much uglier, and the previous version worked fine because PS knows what to do when the line ends without the second half of a binary operator, so the parentheses seems unnecessary.

sarahelsaig (changes requested):

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
BenedekFarkas: I was on the fence about this... PSUseConsistentIndentation detects this (without parenthesis) as incorrect indentation and would remove the indentations so that subsequent lines (that are still part of the string concatenation) start on the same column.

I'm not exactly sure why, but it might have something to do with how it detects whether you're still modifying the positional parameter or concatenating the output of the command (this is just a variable assignment, but the same thing happens when you use string concatenation for the positional parameter of a command, regardless of whether the command has output or not).

The assignment also needs to start on the same line, which is not ideal, but I think it would be a shame not to enable this rule (because it can help a lot) just because of this. Also see https://stackoverflow.com/questions/26372360/powershell-script-indentation-for-long-strings, but I didn't like any of those options.

0liver: Why does the assignment have to start on the same line?

I agree that the previous version is more readable and ... sane.

sarahelsaig: I feel like this does more harm than good. Indentation problems are really easy to notice with the human eye during development and review, so I'm against making such drastic code style compromises just to please the machine.

> Also see https://stackoverflow.com/questions/26372360/powershell-script-indentation-for-long-strings, but I didn't like any of those options.

I like the  ["Binary -join operator"](https://stackoverflow.com/questions/26372360/powershell-script-indentation-for-long-strings#:~:text=Binary%20%2Djoin%20operator) option. It's neat, easy to understand and doesn't suffer from significant whitespace (unlike the unary version or the here-string).

BenedekFarkas: > Why does the assignment have to start on the same line?

If it doesn't, then only second line is indented, the following ones start on the same column as the first one (which, I think is a mistake, but anyway...).

---

I agree with that sentiment Dávid, at this point it's about choosing a lesser evil. The "Binary -join operator" is one of those alternatives that I felt somewhat good about.
Although in terms of syntax (i.e., what you need to change for it work) it's a "superset" of what the code above does.
But it still looks cleaner and is a reasonable compromise, so I'll go ahead and change it, then. BTW the advantage of this one is that you don't need to care about correctly and consistently placing a space character at the beginning or end of a line.

There are cases already (at least in Infrastructure Scripts) using the above solution, so changing those to "Binary -join operator" will be covered by https://github.com/Lombiq/Infrastructure-Scripts/issues/8.

BenedekFarkas: Let me know if these changes are good to go and I'll add them to our PS conventions.

sarahelsaig (approved):

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
BenedekFarkas: Unfortunately, this only works with variable assignments with this structure. When using to display the warning:

```
Write-Warning @(
    'Unable to detect Invoke-ScriptAnalyzer and failed to install PSScriptAnalyzer. If you are on Windows'
    'Powershell, open an administrator shell and type "Install-Module -Name PSScriptAnalyzer -Force'
    "-RequiredVersion $installVersion`". Otherwise see $infoUrl to learn more."
) -join ' '
```
fails, because the array is used as the message parameter before the join operator is applied.

That can be fixed with
```
Write-Warning (@(
        'Unable to detect Invoke-ScriptAnalyzer and failed to install PSScriptAnalyzer. If you are on Windows'
        'Powershell, open an administrator shell and type "Install-Module -Name PSScriptAnalyzer -Force'
        "-RequiredVersion $installVersion`". Otherwise see $infoUrl to learn more."
    ) -join ' ')
```
but then we're back to the original problem (the double-indentation is due to auto-formatting).

Alternative is
```
@(
    'Unable to detect Invoke-ScriptAnalyzer and failed to install PSScriptAnalyzer. If you are on Windows'
    'Powershell, open an administrator shell and type "Install-Module -Name PSScriptAnalyzer -Force'
    "-RequiredVersion $installVersion`". Otherwise see $infoUrl to learn more."
) -join ' ' | Write-Warning
```
but I think it hurts readability.

0liver: Will the following cooperate with auto-formatting?

```pwsh
Write-Warning (
    'Unable to detect Invoke-ScriptAnalyzer and failed to install PSScriptAnalyzer. If you are on Windows' +
    'Powershell, open an administrator shell and type "Install-Module -Name PSScriptAnalyzer -Force' +
    "-RequiredVersion $installVersion`". Otherwise see $infoUrl to learn more."
)
```
Note the `+` signs at the end of the lines.

This would be more than OK with me.

BenedekFarkas: Yes, it works, and this is what I suggested initially - works the same way regardless of assigning the concatenated string or using its value as a positional parameter.

sarahelsaig: I like the `@(...) -join ' ' | Write-Warning`. It's very natural for shell scripting and I don't think it's bad for readability, just unusual for programmers.

As for Oliver's suggestion, my main complaint with your original change was that you placed the first string fragment on the same line as the `(`. If we can avoid that then I'm fine with either solution.

BenedekFarkas: OIC! It looked like to me that you're completely against that format. :D
Yes, what you described works too!

0liver: > my main complaint with your original change was that you placed the first string fragment on the same line as the `(`. If we can avoid that then I'm fine with either solution.

For me too!

I would refrain from using `@(...) -join ' ' | Write-Warning`.

0liver: And most of all, we can use the same formatting for assignment and these statements here, which is wonderful.

BenedekFarkas: I agree that using the same format would be best, but the -join method looks superior to me, because it seems the easiest to maintain: don't need to care about concatenation and placing a space at the beginning or end of a line.

I think I changed my mind on piping to a command (e.g., Write-Warning) - it takes some getting used to, but it's not really harder to read and it also ticks the checkbox for using the same format for variable assignments and positional parameters (it just has a little extra, but its auto-formatting is also the same).

I've just committed those changes, please comment (here):
1. Write-* positional parameter: https://github.com/Lombiq/PowerShell-Analyzers/pull/20/files#diff-dbead90c57f2a0f4701a6796fd07db47b016ea1cf6c7ba24d9f5e949dbbf44cbR102
2. Value assignment: https://github.com/Lombiq/PowerShell-Analyzers/pull/20/files#diff-bc3792324f7234b0bf953b2e9792d3c1b19810bbf16b676d91faaf14aa669328R54

0liver: > I've just committed those changes, please comment (here):

I misread this as:

> I've just committed those changes, please comment (**there**):

My apologies.

0liver: I find the version without `-join ' '` more readable.

In the second example above, this also allows to move the closing parenthesis into the previous line, and saves a line.

Most of all, though, I think the syntax without `-join` is easier to grasp for laymen and beginners.

Here's the obligatory example:

```pwsh
# Sample 1
$message = (
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this' +
    ' variable for consistency.')
```
Over:
```pwsh
# Sample 2
$message = $(
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this'
    'variable for consistency.'
) -join ' '
```

AND:

```pwsh
# Sample 3
Write-Warning (
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this' +
    ' variable for consistency.')
```
Over:
```pwsh
# Sample 4
$(
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this'
    'variable for consistency.'
) -join ' ' | Write-Warning
```

BenedekFarkas: Without trying to sound elitist, I don't think we should worry about the -join method being a bit more advanced than simple concatenation. Using such operators is a basic concept of PS, -join is just one of a dozen or so that's worth using on a frequent basis.

My reasons to use join and this specific syntax:
1. Getting rid of concatenation - this is my primary goal here.
2. Getting rid of space and the beginning or end of a line - I find this also desirable.
3. Placing the supporting syntax in different line (i.e., the opening `@(` in the first line just in itself and closing `) -join ' '` on a new line) separates the text (which what matters primarily) from the supporting syntax.
This, combined with (1) and (2) means that the text (which is the star of the show) is clearly separated from everything else, thus making it easier to read.

Now, I understand that it might look like overkill for just two lines of text, but as soon as just need to update it and/or rewrap it, not to mention expand the text to more lines, it becomes a lot easier to maintain, because the supporting structure is already there, and you only need to care about text lines (not the syntax).
If we can create a snippet for it, then creating such a structure will become trivial too.

0liver: Thanks, you're argumentation is working 😁 I can follow all of these, at the reasoning behind the points you make is sound. Let's do it.

I would suggest, though, that we keep things like `Write-Warning` in front of the message, so it stays clear what's happening here. Having it in the Nth line, at the end, after a pipe, looks like a good place to be overseen. So:


```pwsh
# Sample 3
Write-Warning $(
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this'
    'variable for consistency.'
) -join ' ' 
```
Over:
```pwsh
# Sample 4
$(
    '"$_" is an alias of the automatic variable "$PSItem". Please use the full name of this'
    'variable for consistency.'
) -join ' ' | Write-Warning
```

BenedekFarkas: Thanks! :)

Sample 3 doesn't work (otherwise I'd 100% prefer this layout instead), see: https://github.com/Lombiq/GitHub-Actions/pull/179#discussion_r1072078132, that's why I'd stick to Sample 4 in this case.

0liver: Oh, jolly-bolly!

Oh-boy

Well, then. Let it be.

sarahelsaig (approved):

PR #191, opened 2023-01-18
OSOE-505: Changing the condition against the pull_request event

dorraj07:
this is for testing if the changes were effective or not .


dorraj07:
Closing for testing purposes

PR #194, opened 2023-01-19
OSOE-505: Changing the condition against the pull_request event

### LINKED ISSUE #152: Only run add-jira-issue-code-to-pull-request when the PR is opened (OSOE-505)
Currently, `add-jira-issue-code-to-pull-request` runs when a pull request is opened, and on every subsequent push as well (because `validate-pull-request` and `validate-submodule-pull-request`, where it is executed from, need these triggers). Change this to only run when the PR is freshly opened.

Note that for `validate-pull-request` we use the following triggers:

```
  push:
  pull_request:
    types: [opened, synchronize]
```

Checking on `opened` for this is simple from the `github.event` context. However, `validate-submodule-pull-request` needs to run on every push as well (so it can check for a superproject PR if it wasn't there when the PR was opened), so the only change needed here is to update
https://github.com/Lombiq/GitHub-Actions/blob/20dddc8729dbe1030f734a9e382eb8266e244566/.github/workflows/validate-submodule-pull-request.yml#L24 to `if: github.event.pull_request == opened`

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-505)

dorraj07:
[OSOE-505](https://lombiq.atlassian.net/browse/OSOE-505)
Fixes #152

[OSOE-505]: https://lombiq.atlassian.net/browse/OSOE-505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/validate-pull-request.yml
+++ b/.github/workflows/validate-pull-request.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
-        if: github.event.pull_request.opened == true
+        if: ${{ github.event_name == 'pull_request' && github.event.action == 'opened' }}
         uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@dev
         env:
           JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
--- a/.github/workflows/validate-pull-request.yml
+++ b/.github/workflows/validate-pull-request.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
-        if: github.event.pull_request != ''
+        if: github.event.pull_request == opened
         uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@dev
         env:
           JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
--- a/.github/workflows/validate-pull-request.yml
+++ b/.github/workflows/validate-pull-request.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
+      #adding a comment
         if: github.event.pull_request == opened
         uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@issue/OSOE-505
         env:
--- a/.github/workflows/validate-pull-request.yml
+++ b/.github/workflows/validate-pull-request.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - name: Add Jira Issue Code to Pull Request
         if: github.event.pull_request == opened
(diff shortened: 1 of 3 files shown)


.github/workflows/validate-pull-request.yml
```diff
@@ -15,8 +15,9 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
-        if: github.event.pull_request != ''
-        uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@dev
+      #adding a comment
```
wAsnk: Please delete this comment

.github/workflows/validate-pull-request.yml
```diff
@@ -15,8 +15,9 @@ jobs:
...
+        if: github.event.pull_request == opened
+        uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@issue/OSOE-505
```
wAsnk: This change is not needed.

.github/workflows/validate-submodule-pull-request.yml
```diff
@@ -21,8 +21,8 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
-        if: github.event.pull_request != ''
-        uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@dev
+        if: github.event.pull_request == opened
+        uses: Lombiq/GitHub-Actions/.github/actions/add-jira-issue-code-to-pull-request@issue/OSOE-505
```
wAsnk: This change is not needed.

wAsnk (changes requested):

.github/workflows/validate-submodule-pull-request.yml
dorraj07: which one ?  the reference @issue/OSOE-505 or the condition of the pull_request event ?

wAsnk: If you don't see additional line number which would specify the lines that the comment refers to, then you should consider that it always refers to the line right above the comment. So in this case line 25.

dorraj07: done

.github/workflows/validate-pull-request.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Add Jira Issue Code to Pull Request
-        if: github.event.pull_request != ''
+        if: github.event.pull_request == opened
```
Piedone: Shouldn't these be strings?

dorraj07: after checking, I think it should be github.event.pull_request.opened== true

Piedone: It seems, yes, though check https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=Workflows%20and%20actions%20have%20access%20to%20a%20huge%20amount

wAsnk (approved):

PR #198, opened 2023-02-03
CARE-327: Use 64-bit MSBuild version in msbuild build

0liver:
[CARE-327](https://lombiq.atlassian.net/browse/CARE-327)

[CARE-327]: https://lombiq.atlassian.net/browse/CARE-327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/msbuild-and-test.yml
+++ b/.github/workflows/msbuild-and-test.yml
@@ -88,7 +88,7 @@ jobs:
         uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
 
       - name: Build and Static Code Analysis
-        uses: Lombiq/GitHub-Actions/.github/actions/msbuild@issue/CARE-327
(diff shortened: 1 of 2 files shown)


.github/workflows/msbuild-and-test.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
 
       - name: Build and Static Code Analysis
-        uses: Lombiq/GitHub-Actions/.github/actions/msbuild@dev
+        uses: Lombiq/GitHub-Actions/.github/actions/msbuild@issue/CARE-327
```
adelgeiszt: The dev branch should be restored in this case too.

adelgeiszt (changes requested):

PR #203, opened 2023-02-07
OSOE-523: Optimize spell-checking dictionary files across projects

BenedekFarkas:
[OSOE-523](https://lombiq.atlassian.net/browse/OSOE-523)

[OSOE-523]: https://lombiq.atlassian.net/browse/OSOE-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/spelling/reject.txt
```diff
@@ -0,0 +1 @@
+advices
```
0liver: Please also add `informations` right away.

.github/actions/spelling/expect.txt
```diff
@@ -1,180 +0,0 @@
-abcdef
```
Piedone, line 1: Where are these configured now?

Docs/SpellCheckingConfiguration.md
```diff
@@ -0,0 +1,46 @@
+# Spell checking configuration advice
```
Piedone: I'm an ordinary developer working on one of our projects. During build, I get a spelling error that's a false positive. What should I do? Add a shorthand guide about this, with examples (akin to https://github.com/Lombiq/.NET-Analyzers/blob/dev/Lombiq.Analyzers/Docs/UsingAnalyzersDuringDevelopment.md#practices-on-suppressing-a-rule-for-a-given-piece-of-code) to the advice file. Knowing all the intricacies of configuration is great, but in the majority of cases you won't need all of that.

Docs/Workflows.md
```diff
@@ -60,9 +60,7 @@ jobs:
 
 ## Spelling workflow
 
-Checks for spelling mistakes in a repository using the [Check Spelling](https://github.com/marketplace/actions/check-spelling) GitHub Action, proxied by the [`spelling` action](../.github/actions/spelling/action.yml) in this repository, which has [its own documentation](../.github/actions/spelling/advice.md) describing the configuration options and contribution guidelines. This documentation is also displayed automatically in every spell checking report of a pull request.
-
-You can use already existing configuration files by setting the `spell-check-this` parameter to another existing repository. This parameter is needed even if you want to update our dictionary in a custom branch of project consuming this workflow; changing just the workflow's branch from `dev` to your branch won't take any effect, you can leave it as it is.
```
Piedone, line 65: Elaborate on and add examples for the following scenarios with the configuration we recommend as the default:

1. Generic spell-checking in any project.
2. Generic spell-checking for an OC solution using OSOCE submodules.

Docs/Workflows.md
```diff
@@ -80,19 +78,16 @@ jobs:
...
-      # Use this parameter to list the external dictionary files to use, but beware that check-spelling only accepts flat lists of words (so, for example patterns.txt can't be referenced like this). The order doesn't matter, but sorting it alphabetically makes it easier to maintain.
-      extra-dictionaries: |
+      # Use this parameter to list the external dictionary files to use, but beware that check-spelling only accepts flat lists of words (so, for example patterns.txt can't be referenced like this). The order doesn't matter, but sorting it alphabetically makes it easier to maintain. The "cspell" and "lombiq-lgha" prefixes are available by default - see the spelling action for their exact path.
+      additional-dictionaries: |
         cspell:csharp/csharp.txt
-        lombiq:allow.txt
-        lombiq:expect.txt
-      # This is only needed if you want to use any other dictionaries than the ones in this project.
-      spell-check-this: Lombiq/GitHub-Actions@your-custom-branch
+        lombiq-lgha:Lombiq.common.txt
```
Piedone: I'd include `lombiq-lgha:Lombiq.common.txt` in the default config, since if you don't want our defaults, then don't use LGHA but the action directly. Then infrastructure Scripts and Utility Scripts configuration can be simplified.

.github/actions/spelling/action.yml
```diff
@@ -3,56 +3,148 @@ name: Spelling
...
+
+        Merge-SpellCheckingDictionaryFile @mergeParameters
+
+    - name: Merge dictionary source prefixes
+      id: merge-dictionary-source-prefixes
+      shell: pwsh
+      run: |
+        $defaultPrefixes = @{
+          'lombiq-lgha' = 'https://raw.githubusercontent.com/Lombiq/GitHub-Actions/issue/OSOE-523/.github/actions/spelling/dictionaries/'
+          'cspell' = 'https://raw.githubusercontent.com/check-spelling/cspell-dicts/v20220816/dictionaries/'
```
Piedone: Comment when and how these need to be updated.

Piedone (changes requested):

Docs/SpellCheckingConfiguration.md
BenedekFarkas: That's what the "When not to add entries to dictionary files" section is for. I wrote this documentation "organically" in the sense that I kept refining it based on the situations I encountered across all the projects we applied spell checking to.

I've added everything I think can help but do note that you also have to know about the configuration files to some extent, because addressing a false positive can be done in different ways. That's why this document says at the beginning that you need to be familiar with the configuration options first.

Can you give concrete examples where you wouldn't know what to do?

.github/actions/spelling/action.yml
BenedekFarkas: 'lombiq-lgha' will be updated to use the dev branch before merging this PR and 'cspell' has the associated instructions in the actual spelling step below, because that's where the spelling version is updated (and that's what you need to start with).

On a related note though, I'll expand the docs with instructions for using a branch of LGHA other than dev, because the value of 'lombiq-lgha' should be overridden (so that the correct version of e.g., Lombiq.common.txt is applied).

Docs/Workflows.md
BenedekFarkas: > if you don't want our defaults, then don't use LGHA but the action directly

I wouldn't go that far, because LGHA has now other useful features too that the base action doesn't have, even if you don't need `Lombiq.common.txt`, but having it as a default is indeed easier for us (and if you really don't need it, you can still override the `default-dictionaries` parameter).

.github/actions/spelling/expect.txt
BenedekFarkas: What was expect.txt before is now split into Lombiq.common.txt and OSOCE's own allow.txt. Actually, since we've now settled on other projects excluding submodules (that are also in OSOCE) from spell checking, I'll split OSOCE's allow.txt into an allow.txt and expect.txt (it was like that for some time in this PR). Docs also to be updated (same as https://github.com/Lombiq/GitHub-Actions/pull/203#discussion_r1112285185).

Docs/Workflows.md
BenedekFarkas: Both of those should start with what's in Workflows.md there (minus the line for Lombiq.common.txt, because https://github.com/Lombiq/GitHub-Actions/pull/203#discussion_r1112286436, Workflow.md to be updated with that), i.e., you should start from the absolute minimum configuration not to produce false negatives.
The rest depends on what comes up in the spell-checking report, at which point the contents of SpellCheckingConfiguration.md should be followed.

Docs/SpellCheckingConfiguration.md
Piedone: Currently, you have to read and understand the full page to configure anything, it's too much for the everyday case (as opposed to e.g. when you're setting up a new repo). So, in the advice file I'd add something along the lines of:

> Is this a false positive and there's no actual spelling error? If the word is used only at most in a handful of cases, use [notes on ignore-line]. If it's used repeatedly all over the project, you'll need to extend the corresponding dictionary. For more info on how to configure spell-checking in-depth, check out [link].

BenedekFarkas: That's a good addition.
But an everyday case can be more involved than just adding it to allow.txt, that's why these instructions are important. Whoever changes any spelling configuration file should read SpellCheckingConfiguration.md at least once to be aware of the options, otherwise you can end up with a ton of garbage dictionary entries (which happened to us).

Piedone: If the dictionary needs to extended, people are to be instructed to read that. This was my intent with the above text, but feel free to adjust it as necessary.

Docs/Workflows.md
Piedone: This is important then, please add the approach to this section

Docs/SpellCheckingConfiguration.md
BenedekFarkas: Added your quick tips to advice.md, which already had a CTA to read SpellCheckingConfiguration.md.

Docs/Workflows.md
```diff
@@ -62,7 +62,22 @@ jobs:
...
 
-Example _check-spelling.yml_:
+### How to integrate spell checking into a project
+
+1. Start by adding a job to an existing workflow or create one specific to spell checking with the minimum configuration.
+2. Push the changes and open a pull request to have the initial spell-checking report commented to it.
+3. Follow the [spell checking configuration tips](SpellCheckingConfiguration.md) to work through the list of unrecognized entries.
+4. You will probably end up with a few configuration files and some external dictionaries applied, so your workflow might end up looking something like this.
+
+#### Minimum spell checking step configuration example
```
Piedone: Grammarly suggests "spell-checking". I didn't find an authoritative answer for the spelling (haha) of this, please check.

Docs/Workflows.md
```diff
@@ -62,7 +62,22 @@ jobs:
 
 Checks for spelling mistakes in a repository using the [Check Spelling](https://github.com/marketplace/actions/check-spelling) GitHub Action, proxied by the [`spelling` action](../.github/actions/spelling/action.yml) in this repository, which has [its own documentation](SpellCheckingConfiguration.md) describing the configuration options and contribution guidelines. This documentation is also linked automatically at the end of every spell checking report of a pull request.
 
-Example _check-spelling.yml_:
+### How to integrate spell checking into a project
+
+1. Start by adding a job to an existing workflow or create one specific to spell checking with the minimum configuration.
+2. Push the changes and open a pull request to have the initial spell-checking report commented to it.
+3. Follow the [spell checking configuration tips](SpellCheckingConfiguration.md) to work through the list of unrecognized entries.
+4. You will probably end up with a few configuration files and some external dictionaries applied, so your workflow might end up looking something like this.
```
```suggestion
4. You will probably end up with a few configuration files and some external dictionaries applied, so your workflow might end up looking something like the complete example below.
```

Docs/SpellCheckingConfiguration.md
```diff
@@ -4,11 +4,14 @@
...
 
+When integrating spell checking to your project for the first time or working on resolving a greater number of unrecognized entries, take small, incremental steps and don't rush adding everything to a dictionary file. Some entries can signal that certain file types or specific files should be excluded and in some cases a useful ignore pattern can emerge too.
+
 ## General tips
 
+- You can reduce your spell checking configuration by excluding submodules that are spell-checked on their own or as part of a larger parent project. The best examples are [OrchardCore.Commerce](https://github.com/OrchardCMS/OrchardCore.Commerce) and having a submodule that is also a submodule of [Open Source Orchard Core Extensions](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions): Submitting a pull request in such a submodule requires you to open a parent-PR in OSOCE to make sure that spell checking is executed on the updated code before it gets merged.
```
Piedone:
```suggestion
- You can reduce your spell checking configuration by excluding submodules that are spell-checked on their own or as part of a larger parent project. The best examples are [OrchardCore.Commerce](https://github.com/OrchardCMS/OrchardCore.Commerce) and having a submodule that is also a submodule of [Open Source Orchard Core Extensions (OSOCE)](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions): Submitting a pull request in such a submodule requires you to open a parent PR in OSOCE to make sure that spell checking is executed on the updated code before it gets merged.
```

Docs/SpellCheckingConfiguration.md
```diff
@@ -25,22 +28,31 @@ You can provide these files in your own repository and by default they must be p
...
+- Only add one dictionary at a time so that recommendations are re-evaluated correctly. This makes it easier to keep your configuration at the necessary minimum, since some dictionaries are overlapping.
 - Also check if the contents of the dictionary make sense. Some recommended libraries seem to have a great number of entries that seem useless in light of how the spell checking process works now. That is likely due to those dictionary files being outdated and fundamental changes have been made to process since.
-- Try to avoid large dictionaries, like the one for C++: It has over 30 thousand entries, because it casts a net way too large and contains a ton of entries that should have been excluded by other means (e.g., a pattern, ignoring a line or a while file).
+- Try to avoid large dictionaries, like the one for C++: It has over 30 thousand entries, because it casts a net way too large and contains a ton of entries that should have been excluded by other means (e.g., a pattern, ignoring a line or a whole file).
```
Piedone, line 34: Explain why it's good to avoid large dictionaries.

Docs/SpellCheckingConfiguration.md
```diff
@@ -25,22 +28,31 @@ You can provide these files in your own repository and by default they must be p
...
 
-1. _Merge-SpellCheckingDictionaryFile.ps1_: Use this to maintain your _excludes.txt_ file by adding the entries from another file, while still keeping your own. To just remove duplicates and sort the entries alphabetically in a single configuration file, pass in the same file for both parameters.
-2. _Optimize-SpellCheckingDictionaryFile.ps1_: Use this to remove entries from your dictionary files that are already present in an external one you're referencing.
+- _Merge-SpellCheckingDictionaryFile.ps1_: Use this to maintain your _excludes.txt_ file by adding the entries from another file, while still keeping your own. To just remove duplicates and sort the entries alphabetically in a single configuration file, pass in the same file for both parameters.
+- _Optimize-SpellCheckingDictionaryFile.ps1_: Use this to remove entries from your dictionary files that are already present in an external one you're referencing.
+
+## Working with a non-dev branch of Lombiq.GitHub.Actions
```
Piedone:
```suggestion
## Working with a non-dev branch of `Lombiq.GitHub.Actions`
```

Piedone (approved):

Docs/Workflows.md
0liver: Grammarly is right here, because you're using this word pair as an [adjective](https://en.wikipedia.org/wiki/Adjective) to "step configuration".

This is the same as saying "We do open source." vs. "We build open-source software."

Docs/SpellCheckingConfiguration.md
BenedekFarkas: It's explained in this section from multiple angles, so please read the section as a whole, not just individual lines. What's missing for your understanding?

Piedone: I've read it, and I don't see it explaining how specifically _large_ (i.e. not overlapping, or ones causing false negatives; talking about the size, not content) dictionaries cause an issue. Is it a performance issue?

BenedekFarkas: I haven't considered that, but I imagine it affects performance too to some extent. Large means a lot of entries.

Piedone: OK, let's leave it like that.

Piedone (approved):

PR #208, opened 2023-02-13
OSOE-536: In NuGet publish workflow, restrict valid versions to vM.N.O or v.M.N.O-alpha.X.[issue-code]

### LINKED ISSUE #187: In NuGet publish workflow, restrict valid versions to vM.N.O or vM.N.O-alpha.X.[issue-code] (OSOE-536)
[Jira issue](https://lombiq.atlassian.net/browse/OSOE-536)

adelgeiszt:
[OSOE-536](https://lombiq.atlassian.net/browse/OSOE-536)
Fixes #187 

[OSOE-536]: https://lombiq.atlassian.net/browse/OSOE-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -62,7 +62,7 @@ runs:
 
         if ($version -eq 'USE_GITHUB_REF_NAME') {
             $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
-            if (!($version -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
+            if (!($version -match '^v?\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
                 Write-Output "Invalid version number! The correct format is vM.N.O or v.M.N.O-alpha.X.[issue-code]."
                 exit 1
             }
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -61,7 +61,7 @@ runs:
         '@.Trim()
 
         if ($version -eq 'USE_GITHUB_REF_NAME') {
-            if (($version -match '^v?\d+(\.\d+){0,3}-(\w+\.)+\d+\.\w+-\d+$') -or ($version -match '^v?\d+(\.\d+){0,3}$')) {
+            if ($version -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$') {
                 $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
             } else {
                 Write-Output "Invalid version number! The correct format is vM.N.O or v.M.N.O-alpha.X.[issue-code]."
@@ -70,7 +70,7 @@ runs:
         }
         if ($version -eq 'USE_GITHUB_RUN_NUMBER') {
           $tags = git tag |
-            ? { ($_ -match '^v?\d+(\.\d+){0,3}$') -or ($_ -match '^v?\d+(\.\d+){0,3}-(\w+\.)+\d+\.\w+-\d+$') } |
+            ? { ($_ -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$') } |
             % { [version]($_.TrimStart("v")) } |
             Sort-Object -Descending
           $current = if ($tags.Count -gt 0) { $tags[0] } else { [version]"0.0.0" }
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -69,7 +69,7 @@ runs:
         }
         if ($version -eq 'USE_GITHUB_RUN_NUMBER') {
           $tags = git tag |
-            ? { $_ -match '^v?\d+(\.\d+){0,3}$' } |
+            ? { $_ -match '^v?\d+\.\d+\.\d+$' } |
             % { [version]($_.TrimStart("v")) } |
             Sort-Object -Descending
           $current = if ($tags.Count -gt 0) { $tags[0] } else { [version]"0.0.0" }
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -63,7 +63,7 @@ runs:
         if ($version -eq 'USE_GITHUB_REF_NAME') {
             $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
             if (!($version -match '^\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
-                Write-Output "::error::$version is an invalid version number! The correct format is vM.N.O or vM.N.O-alpha.X.[issue-code]."
+                Write-Output "::error::`"v$version`" is invalid. The correct format is `"vM.N.O`" or `"vM.N.O-alpha.X.<issue-code>`"."
                 exit 1
             }
         }
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -60,7 +60,14 @@ runs:
         ${{ inputs.publish-version }}
         '@.Trim()
 
-        if ($version -eq 'USE_GITHUB_REF_NAME') { $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v") }
+        if ($version -eq 'USE_GITHUB_REF_NAME') {
+            if (($version -match '^v?\d+(\.\d+){0,3}-(\w+\.)+\d+\.\w+-\d+$') -or ($version -match '^v?\d+(\.\d+){0,3}$')) {
+                $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+            } else {
+                Write-Output "version number is not in a correct format."
+                exit 1
+            }
+        }
         if ($version -eq 'USE_GITHUB_RUN_NUMBER') {
           $tags = git tag |
             ? { $_ -match '^v?\d+(\.\d+){0,3}$' } |
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -63,7 +63,7 @@ runs:
         if ($version -eq 'USE_GITHUB_REF_NAME') {
             $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
             if (!($version -match '^\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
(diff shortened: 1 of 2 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -60,10 +60,16 @@ runs:
...
+        if ($version -eq 'USE_GITHUB_REF_NAME') {
+            $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+            if (!($version -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
+                Write-Output "Invalid version number! The correct format is vM.N.O or v.M.N.O-alpha.X.[issue-code]."
+                exit 1
+            }
+        }
         if ($version -eq 'USE_GITHUB_RUN_NUMBER') {
           $tags = git tag |
-            ? { $_ -match '^v?\d+(\.\d+){0,3}$' } |
```
0liver, line 66: @DAud-IcI I saw that you've committed the optional `v` at the beginning of the regex [here](https://github.com/Lombiq/GitHub-Actions/commit/3b4998b728a4090ec#diff-5faf43e0bf23a1323690a2c9fb6921bd066a6ae504d8a982fb85ad25bc99e3e6R55), but I would suggest to drop that since we have a policy that version numbers must always have a `v` prefix. Makes the code simpler and avoids surprises, I would argue.

.github/actions/publish-nuget/action.yml
```diff
@@ -60,10 +60,16 @@ runs:
...
+            ? { ($_ -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$') } |
```
0liver: I don't think we should be changing this regex here. In the [description](https://github.com/Lombiq/GitHub-Actions/pull/208/files#diff-5faf43e0bf23a1323690a2c9fb6921bd066a6ae504d8a982fb85ad25bc99e3e6R28) of the `publish-version` parameter it says that in the case of `USE_GITHUB_RUN_NUMBER` we look at the latest **non-prerelease** version.

That was intentional, right @DAud-IcI?

.github/actions/publish-nuget/action.yml
```diff
@@ -60,10 +60,16 @@ runs:
         ${{ inputs.publish-version }}
         '@.Trim()
 
-        if ($version -eq 'USE_GITHUB_REF_NAME') { $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v") }
+        if ($version -eq 'USE_GITHUB_REF_NAME') {
+            $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+            if (!($version -match '^v?(\d+\.\d+\.\d+)(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
```
0liver: `$version` can never start with a `v` since you remove that in the line above.

0liver: You don't need to group the match around the first three digits:
```suggestion
            if (!($version -match '^v?\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
```

.github/actions/publish-nuget/action.yml
0liver, line 66: @DAud-IcI Also, can we agree to simply using `\d+\.\d+\.\d+` for the digits? I don't think we want or need support for 4 digits.

.github/workflows/publish-nuget.yml
```diff
@@ -73,17 +73,17 @@ jobs:
     timeout-minutes: ${{ inputs.timeout-minutes }}
     steps:
       - name: Checkout
-        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@issue/OSOE-536
```
0liver: You only need to change the branch selectors for actions that you've changed as part of your issue.

.github/workflows/publish-nuget.yml
```diff
@@ -73,17 +73,17 @@ jobs:
...
 
       - name: Set up .NET
-        uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
+        uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@issue/OSOE-536
         with:
           dotnet-version: ${{ inputs.dotnet-version }}
 
       - name: Publish to NuGet
-        uses: Lombiq/GitHub-Actions/.github/actions/publish-nuget@dev
+        uses: Lombiq/GitHub-Actions/.github/actions/publish-nuget@issue/OSOE-536
```
0liver, line 87: This is the only branch reference you need to change for this issue.

.github/actions/publish-nuget/action.yml
0liver: Add the version information to the parameter description.

0liver (changes requested):

.github/actions/publish-nuget/action.yml
sarahelsaig: sure

sarahelsaig: I think we ought to support any [Version](https://learn.microsoft.com/en-us/dotnet/api/system.version?view=net-7.0) string. Why would we make it stricter than standard?

.github/actions/publish-nuget/action.yml
sarahelsaig: Yes, please don't change it! This line intentionally only looks for release tags to determine the current release version. This number is incremented to get the new prerelease version. It is used for automatic prerelease on merge. So alpha shouldn't be allowed here, or you get errors:
image

adelgeiszt: Ok, thank you. I restored the original regex here.

.github/actions/publish-nuget/action.yml
```diff
@@ -60,7 +60,13 @@ runs:
         ${{ inputs.publish-version }}
         '@.Trim()
 
-        if ($version -eq 'USE_GITHUB_REF_NAME') { $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v") }
+        if ($version -eq 'USE_GITHUB_REF_NAME') {
+            $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+            if (!($version -match '^\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
+                Write-Output "Invalid version number! The correct format is vM.N.O or vM.N.O-alpha.X.[issue-code]."
```
sarahelsaig: Should we use `{issue-code}` instead of `[issue-code]`? For people used to shell scripting [something] suggests that it's optional. We also use braces in string interpolation so it's a more natural fit.

sarahelsaig: Please make it a [workflow command](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message) (so it generates an error annotation) by adding `::error::` at the beginning of the string.

Also I'd include the current value of `$version` in the message so the problem is very clear on the summary page too.

sarahelsaig (changes requested):

.github/actions/publish-nuget/action.yml
0liver: I was thinking that the issue code could be **optional**, that's why I uses square brackets in the issue description. But no-one asked for that, so let's not complicate things further.

0liver: Braces hide the value in GHA output: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#masking-a-value-in-log.

.github/actions/publish-nuget/action.yml
0liver: In the context of **NuGet** packages, I wouldn't call the .NET `Version` class our standard. We'd rather need to look here: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics.

Since we're going to use only three digits in the resulting version, too (see [line 77 below](https://github.com/Lombiq/GitHub-Actions/pull/208/files/be54d65b0df4efa330b52ed3bfaa39fc9f483dcb#diff-5faf43e0bf23a1323690a2c9fb6921bd066a6ae504d8a982fb85ad25bc99e3e6R77)), I think it would be misleading to allow 4 digits as the input.

.github/actions/publish-nuget/action.yml
0liver: ~Please remove `v?` from the beginning of the regex.~

Nevermind.

.github/actions/publish-nuget/action.yml
```diff
@@ -60,7 +60,13 @@ runs:
         ${{ inputs.publish-version }}
         '@.Trim()
 
-        if ($version -eq 'USE_GITHUB_REF_NAME') { $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v") }
+        if ($version -eq 'USE_GITHUB_REF_NAME') {
+            $version = $Env:GITHUB_REF_NAME.Trim().TrimStart("v")
+            if (!($version -match '^\d+\.\d+\.\d+(-alpha\.\d+\.[a-z]{2,}-\d+)?$')) {
+                Write-Output "::error::$version is an invalid version number! The correct format is vM.N.O or vM.N.O-alpha.X.[issue-code]."
```
0liver: We could use angle brackets instead of the optional-looking square brackets then, @DAud-IcI:

```suggestion
                Write-Output "::error::`"v$version`" is invalid. The correct format is `"vM.N.O`" or `"vM.N.O-alpha.X.<issue-code>`"."
```

0liver: Please also change it in the documentation.

.github/workflows/publish-nuget.yml
```diff
@@ -55,8 +55,8 @@ on:
         type: string
         default: "USE_GITHUB_REF_NAME"
         description: >
-          The desired NuGet package version used for publishing. If not specified, the GITHUB_REF_NAME environment
-          variable is used which is suitable if the version is derived from a git tag.
+          The desired NuGet package version used for publishing. The correct input format is vM.N.O or vM.N.O-alpha.X.[issue-code].
+          If not specified, the GITHUB_REF_NAME environment variable is used which is suitable if the version is derived from a git tag.
```
0liver: Please adhere to 120 characters per line.

.github/actions/publish-nuget/action.yml
sarahelsaig: Agree.

.github/actions/publish-nuget/action.yml
0liver: @DAud-IcI ?

sarahelsaig: okay then

0liver: @adelgeiszt Please use `\d+\.\d+\.\d+` here, then.

0liver (approved):

PR #222, opened 2023-03-30
OSOE-566: Fixing that the publish-nuget workflow doesn't fail when a package build has an error

### LINKED ISSUE #207: The publish-nuget workflow doesn't fail when a package build has an error. (OSOE-566)
Here is an example: https://github.com/Lombiq/Orchard-Vue.js/actions/runs/4146253609

I believe the error was thrown by the `dotnet pack`, but it's not sure from the log. In the end out of the 3 projects in Lombiq.VueJs only two were uploaded to NuGet. Since the workflow didn't fail, I didn't get a warning email so I could only find out about the problem when I dug into the run logs.

(I have already fixed that specific problem with Lombiq.VueJs, so this issue is not a blocker.)

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-566)

Piedone:
[OSOE-566](https://lombiq.atlassian.net/browse/OSOE-566)
Fixes #207

[OSOE-566]: https://lombiq.atlassian.net/browse/OSOE-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -34,7 +34,7 @@ foreach ($project in $projects)
 
     if ($LASTEXITCODE -ne 0)
     {
-        Write-Output "::error::dotnet pack failed for the project $($project.FullName)."
+        Write-Error "::error::dotnet pack failed for the project $($project.FullName)."
         exit 1
     }
 
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -34,7 +34,7 @@ foreach ($project in $projects)
 
     if ($LASTEXITCODE -ne 0)
     {
-        Write-Error "dotnet pack failed for the project $($project.FullName)."
+        Write-Output "::error::dotnet pack failed for the project $($project.FullName)."
         exit 1
     }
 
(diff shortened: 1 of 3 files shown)


.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -32,5 +32,11 @@ foreach ($project in $projects)
         dotnet pack $project @Arguments
     }
 
+    if ($LASTEXITCODE -ne 0)
+    {
+        Write-Error "dotnet pack failed for the project $($project.FullName)."
```
sarahelsaig: Shouldn't this be an error annotation?
```suggestion
        Write-Error "::error::dotnet pack failed for the project $($project.FullName)."
```

sarahelsaig (changes requested):

.github/actions/publish-nuget/New-NuGetPackage.ps1
Piedone: Yep, new run: https://github.com/Lombiq/Orchard-Vue.js/actions/runs/4567982143 Note that I've used `Write-Output`, as we did in similar cases before. The result is that the "dotnet pack failed for the project /home/runner/work/Orchard-Vue.js/Orchard-Vue.js/Lombiq.VueJs/Lombiq.VueJs.csproj." error also appears under a run summary vs [not with `Write-Error`](https://github.com/Lombiq/Orchard-Vue.js/actions/runs/4567982143/attempts/5).

PR #223, opened 2023-04-03
OSOE-593: Add job summary with tool versions

0liver:
[OSOE-593](https://lombiq.atlassian.net/browse/OSOE-593)

[OSOE-593]: https://lombiq.atlassian.net/browse/OSOE-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/print-config-summary/action.yml
+++ b/.github/actions/print-config-summary/action.yml
@@ -7,6 +7,6 @@ runs:
     - name: Write configuration to summary
       shell: pwsh
       run: |
-        "dotnet: $(dotnet --version)" >> $GITHUB_STEP_SUMMARY
-        "node  : $(node -v)" >> $GITHUB_STEP_SUMMARY
-        "pnpm  : $(pnpm -v)" >> $GITHUB_STEP_SUMMARY
+        "### dotnet: $(dotnet --version)" >> $GITHUB_STEP_SUMMARY
+        "### node  : $(node -v)" >> $GITHUB_STEP_SUMMARY
+        "### pnpm  : $(pnpm -v)" >> $GITHUB_STEP_SUMMARY
--- a/.github/actions/print-config-summary/action.yml
+++ b/.github/actions/print-config-summary/action.yml
@@ -1,0 +1,12 @@
+name: Print summary of important configuration
+description: Contains .NET, Node.js, and pnpm versions, among others
+
+runs:
+  using: "composite"
+  steps:
+    - name: Write configuration to summary
+      shell: pwsh
+      run: |
+        "dotnet: $(dotnet --version)" >> $GITHUB_STEP_SUMMARY
+        "node  : $(node -v)" >> $GITHUB_STEP_SUMMARY
+        "pnpm  : $(pnpm -v)" >> $GITHUB_STEP_SUMMARY
--- a/.github/actions/print-config-summary/action.yml
+++ b/.github/actions/print-config-summary/action.yml
@@ -7,6 +7,11 @@ runs:
     - name: Write configuration to summary
       shell: pwsh
       run: |
-        "### dotnet: $(dotnet --version)" >> $env:GITHUB_STEP_SUMMARY
-        "### node  : $(node -v)" >> $env:GITHUB_STEP_SUMMARY
-        "### pnpm  : $(pnpm -v)" >> $env:GITHUB_STEP_SUMMARY
+        $toolVersions = @"
+          | Tool | Version |
+          | ---- | ------- |
+          | dotnet | $(dotnet --version) |
+          |  node  | $(node -v) |
+          |  pnpm  | $(pnpm -v) |
+        "@
+        $toolVersions >> $env:GITHUB_STEP_SUMMARY
--- a/.github/actions/print-config-summary/action.yml
+++ b/.github/actions/print-config-summary/action.yml
@@ -9,9 +9,9 @@ runs:
       run: |
         $toolVersions = @"
           | Tool | Version |
-          | ---- | ------- |
+          | :--- | ------: |
           | dotnet | $(dotnet --version) |
-          |  node  | $(node -v) |
-          |  pnpm  | $(pnpm -v) |
+          | node | $(node -v) |
+          | pnpm | $(pnpm -v) |
         "@
         $toolVersions >> $env:GITHUB_STEP_SUMMARY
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -213,6 +213,10 @@ jobs:
           binary-log-artifact-retention-days: ${{ inputs.build-binary-log-artifact-retention-days }}
           warnings-as-errors: ${{ inputs.warnings-as-errors }}
 
+      - name: Print configuration summary
+        if: always()
```
dministro: Do we need this report if the workflow is canceled?
If not:
```suggestion
        if: success() || failure()
```

.github/actions/print-config-summary/action.yml
```diff
@@ -0,0 +1,17 @@
...
+  steps:
+    - name: Write configuration to summary
+      shell: pwsh
+      run: |
+        $toolVersions = @"
+          | Tool | Version |
+          | ---- | ------- |
+          | dotnet | $(dotnet --version) |
+          |  node  | $(node -v) |
+          |  pnpm  | $(pnpm -v) |
```
dministro: Single space around cell content and three hyphens are enough.

With alignment:
```suggestion
          | Tool | Version |
          | :--- | ---: |
          | dotnet | $(dotnet --version) |
          | node | $(node -v) |
          | pnpm | $(pnpm -v) |
```

dministro (changes requested):

.github/workflows/build-and-test-orchard-core.yml
0liver: Yes, we want the report most of all in the case of an error 😁

.github/actions/print-config-summary/action.yml
0liver: I liked my version 😁

Reflecting on this, I would prefer to have the separator row with the same column widths as the header, but I can agree that in the table body the alignment of the cells might not be as important.

I like the left/right alignment!

0liver: So, how about:

```suggestion
          | Tool | Version |
          | :--- | ------: |
          | dotnet | $(dotnet --version) |
          | node | $(node -v) |
          | pnpm | $(pnpm -v) |
```

.github/workflows/build-and-test-orchard-core.yml
dministro: The `always()` condition is evaluated as `true` in case of manual cancellation or cancellation by another workflow while `success() || failure()` is not.
See related: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#Conventions

.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -213,6 +213,10 @@ jobs:
...
+        uses: Lombiq/GitHub-Actions/.github/actions/print-config-summary@issue/OSOE-593
```
dministro: I would suggest making this step optional with the default `false` and enabled in the super project workflow, since possible this is used by other projects.

.github/workflows/build-and-test-orchard-core.yml
0liver: I am aware of that, and like a said, we want this info to show up **specifically** when a workflow job fails (which then cancels it).

.github/workflows/build-and-test-orchard-core.yml
0liver: Let's not overdo this right now. There's no harm in showing that info - and actually I would like it to be enabled by default so that I can access the version info wherever the workflow is run.

We can make it configurable to disable later, but right now, let's please merge it.

.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -213,6 +213,10 @@ jobs:
           binary-log-artifact-retention-days: ${{ inputs.build-binary-log-artifact-retention-days }}
           warnings-as-errors: ${{ inputs.warnings-as-errors }}
 
+      - name: Print configuration summary
```
Piedone, line 216: Is this really generally necessary for every build? Because if not, then have it configurable, disabled by default, and only enabled for OSOCE.

0liver: As I said above (https://github.com/Lombiq/GitHub-Actions/pull/223#discussion_r1157306921), the NE builds have been failing in different projects, and it would be cumbersome to have them all update their workflow configuration right now.

This is still work in progress, and I'm trying to gather the version data to see if the used versions of Node.js and PNPM are connected to the build errors.

Once we find out more and won't need this by default, we can definitely turn it off. But not at the moment.

.github/workflows/build-and-test-orchard-core.yml
dministro: Noted this under the [issue](https://lombiq.atlassian.net/browse/OSOE-593).
@Piedone @0liver

.github/workflows/build-and-test-orchard-core.yml
dministro: You say, we need this information if eg the [spelling](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/workflows/spelling.yml) workflow failed?

I'm sure, the `Print configuration summary` step will be executed in case the `Build and Static Code Analysis` get failed if you are using `success() || failure()` condition here.
See condition in [Cancel Workflow on Failure](https://github.com/Lombiq/GitHub-Actions/blob/30aacfb73e7bb869cf83063547d0344ff0ae44e8/.github/workflows/build-and-test-orchard-core.yml#L240)

https://github.com/Lombiq/GitHub-Actions/blob/30aacfb73e7bb869cf83063547d0344ff0ae44e8/.github/workflows/build-and-test-orchard-core.yml#L239-L243

.github/workflows/build-and-test-orchard-core.yml
dministro: @Piedone I noted this under the issue. Can you accept this as temporary?

Piedone: OK

dministro (approved):

PR #228, opened 2023-04-28
LMBQ-101: LGHA has "Precompile Orchard 1 application" action

BenedekFarkas:
[LMBQ-101](https://lombiq.atlassian.net/browse/LMBQ-101)

[LMBQ-101]: https://lombiq.atlassian.net/browse/LMBQ-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/precompile-orchard1-app/action.yml
```diff
@@ -0,0 +1,85 @@
+name: Precompile Orchard 1 application
+description: Publishes an Orchard 1 application using the Precompiled build target through Orchard.proj.
```
Piedone, line 2: If it's intentionally not documented, add "Intentionally not documented in Actions.md since it's only meant for internal use."

.github/actions/precompile-orchard1-app/action.yml
```diff
@@ -0,0 +1,85 @@
...
+  solution-path:
+    required: true
+    description: The path to the solution file, relative to Orchard.proj.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Clone repository
+      if: inputs.repository != ''
+      uses: actions/checkout@v3.1.0
```
Piedone, line 58: Please document why not our `checkout` action is used (it wouldn't make sense due to this being a legacy workflow that's the only one needing more parameters).

Piedone (changes requested):

.github/actions/precompile-orchard1-app/action.yml
BenedekFarkas: It was intentional, but it can actually be used to simplify two workflows in a client project we've recently migrated to GHA, so I'll document it instead.

Piedone (approved):

PR #232, opened 2023-05-26
HAST-175: NuGet publishing fixes: Proper MSBuild property checks and safer SourceLink package add

Piedone:
[HAST-175](https://lombiq.atlassian.net/browse/HAST-175)

[HAST-175]: https://lombiq.atlassian.net/browse/HAST-175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,21 +16,70 @@
...
+        [string] $PropertyName
+    )
+
+    try
+    {
+        $projectFileContent = Get-Content $ProjectFilePath -ErrorAction Stop
+
+        $newTarget = @"
+  <Target Name="GetPropertyValue">
+    <Message Importance="High" Text="`$($PropertyName)" />
```
sarahelsaig: Just to be safe, it may be good to surround it with some delimiter such as
```xml
<Message Importance="High" Text="---Get-ProjectProperty---`$($PropertyName)---Get-ProjectProperty---" />
```
and then trim it out like 
```pwsh
$buildOutput = (dotnet msbuild ...).Split("---Get-ProjectProperty---")[1]`
```

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,21 +16,70 @@
...
+  </Target>
+"@
+
+        # Insert the new target XML string just before the closing </Project> tag.
+        $updatedProjectFileContent = $projectFileContent -replace '</Project>', "$newTarget`r`n</Project>"
+
+        # Write the updated content back to the project file.
+        Set-Content $ProjectFilePath $updatedProjectFileContent -ErrorAction Stop
```
sarahelsaig: Instead of overwriting and then having to restore the original project file, why not write into a new copy and then delete it? e.g.
```pwsh
$Extension = (Get-Item $ProjectFilePath).Extension
$TemporaryProjectFilePath = $ProjectFilePath -replace "\$Extension`$", ".GetProperty$Extension"
Set-Content $TemporaryProjectFilePath $updatedProjectFileContent -ErrorAction Stop
...
Remove-Item $TemporaryProjectFilePath
```

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,21 +16,70 @@
...
+
+        $buildOutput = dotnet msbuild $ProjectFilePath /nologo /v:minimal /p:DesignTimeBuild=true /p:BuildProjectReferences=false /t:GetPropertyValue
+
+        # Restore the file content.
+        Set-Content $ProjectFilePath $projectFileContent -ErrorAction Stop
+
+        Write-Output ([string]::IsNullOrEmpty($buildOutput) ? '' : $buildOutput.Trim())
```
sarahelsaig: Is outputting an empty string necessary here? If not, then it feels like an `if` would make more sense than a ternary here.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,21 +16,70 @@
...
+    }
+    catch
+    {
+        Write-Error "Failed to add the GetPropertyValue target: $($_.Exception.Message)."
```
sarahelsaig: May we prepend `::error::` here?

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -16,21 +16,70 @@
...
 $projects = (Test-Path *.sln) ? (dotnet sln list | Select-Object -Skip 2 | Get-Item) : (Get-ChildItem *.csproj)
 
 foreach ($project in $projects)
 {
-    $projectFile = [xml](Get-Content $project)
-    $isPackable = $projectFile.SelectSingleNode('//PropertyGroup/IsPackable').InnerText
+    Write-Output "Packing $($project.Name)..."
+
+    $isPackableProperty = Get-ProjectProperty -ProjectFilePath  $project -PropertyName 'IsPackable'
+    $isPackable = [string]::IsNullOrEmpty($isPackableProperty) -or $isPackableProperty -eq 'true'
```
sarahelsaig: So `$isPackableProperty -ne 'false'`? Or rather `$isPackableProperty -notlike '*false*'` for the sake of case insensitivity and whitespace tolerance? I think that fits better with the text of the warning below too.

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -28,24 +28,39 @@ else
...
 
 $tests = dotnet sln $Solution list |
     Select-Object -Skip 2 |
     Select-String '\.Tests\.' |
     Select-String -NotMatch 'Lombiq.Tests.UI.csproj' |
     Select-String -NotMatch 'Lombiq.Tests.csproj' |
     Where-Object {
-        $result = dotnet test $optOut --configuration $Configuration --list-tests --verbosity $Verbosity $PSItem 2>&1 | Out-String -Width 9999
-        -not [string]::IsNullOrEmpty($result) -and $result.Contains('The following Tests are available')
+        # While the test projects are run individually, passing in the solution name via the conventional MSBuild #
```
sarahelsaig: Is the `#` at the end of this line intentional or a typo?

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -28,24 +28,39 @@ else
...
+        # property allows build customization.
+        $switches = @(
+            "--configuration:$Configuration"
+            '--list-tests'
+            "--verbosity:$Verbosity"
+            "-p:SolutionName=""$solutionName"""
+        )
+        # Without Out-String, Contains() below won't work for some reason.
```
sarahelsaig: Should we have an empty line above each of these two comments?

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -28,24 +28,39 @@ else
...
+        $output = dotnet test @switches $PSItem 2>&1 | Out-String -Width 9999
+
+        if ($LASTEXITCODE -ne 0)
+        {
+            $errorMessage = "dotnet test failed for the project $PSItem with the following output:`n$output"
+            throw $errorMessage
```
sarahelsaig: Is `throw` appropriate here? There is no corresponding try-catch, so probably not. How about this?
```pwsh
Write-Error "::error::$errorMessage"
exit 1
```

sarahelsaig (changes requested):

.github/actions/publish-nuget/New-NuGetPackage.ps1
Piedone: Outputting it is not necessary, but we need to return something from the function, so I don't see why an `if` would be better.

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
Piedone: Before this one yes, but the first one is just after the opening brace, where there shouldn't really be an empty line.

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: I meant instead of the whole expression. Once you check that it's not `false` you don't need to separately check if it's null or empty:
image

PR #237, opened 2023-06-07
SPAL-23: Update default .NET version in workflows, prevent version jumps

0liver:
[SPAL-23](https://lombiq.atlassian.net/browse/SPAL-23)

[SPAL-23]: https://lombiq.atlassian.net/browse/SPAL-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -142,7 +142,10 @@ runs:
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
-        "{ ""sdk"": { ""version"": ""$Env:DOTNET_SDK_TO_USE"" } }" > ./global.json
+        if (-not (Test-Path -Path .\global.json -PathType Leaf))
+        {
+          "{ ""sdk"": { ""version"": ""$Env:DOTNET_SDK_TO_USE"" } }" > .\global.json
+        }
 
     - name: Build and Static Code Analysis
       id: build
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -159,7 +159,7 @@ runs:
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
-        Write-Output "Using .NET SDK from ./global.json:`n$(Get-Content .\global.json)"
+        Write-Output "Using .NET SDK from ./global.json:`n$(Get-Content .\global.json -Raw)"
 
         Initialize-ArtifactNameSuffix '${{ inputs.directory }}'
 
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -144,7 +144,7 @@ runs:
       run: |
         $globalJsonExists = Test-Path -Path .\global.json -PathType Leaf
         # Read from file or use empty object.
-        $globalJsonContent = if ($globalJsonExists) { Get-Content .\global.json -Raw } else { '{}' }
+        $globalJsonContent = $globalJsonExists ? (Get-Content .\global.json -Raw) : '{}'
         $globalJsonObject = $globalJsonContent | ConvertFrom-Json
         # Only set the sdk property if it does not exist yet.
         if (-not (Get-Member -InputObject $globalJsonObject -Name 'sdk' -MemberType NoteProperty)) {
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -137,7 +137,8 @@ runs:
         key: ${{ steps.detect-cache-configuration.outputs.key }}
         restore-keys: ${{ steps.detect-cache-configuration.outputs.restore-keys }}
 
-    - name: Create temporary global.json with desired .NET SDK version
+    - name: Create Temporary global.json with Desired .NET SDK Version
+      # We do this to force `dotnet build` to use that exact version and avoid breaking builds by silent upgrades.
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
--- a/.github/actions/build-dotnet/action.yml
+++ b/.github/actions/build-dotnet/action.yml
@@ -141,7 +141,7 @@ runs:
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
-        echo '{ "sdk": { "version": "$Env:DOTNET_SDK_TO_USE" } }' > ./global.json
+        "{ ""sdk"": { ""version"": ""$Env:DOTNET_SDK_TO_USE"" } }" > ./global.json
 
     - name: Build and Static Code Analysis
       id: build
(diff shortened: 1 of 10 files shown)


.github/actions/build-dotnet/action.yml
```diff
@@ -137,11 +137,19 @@ runs:
         key: ${{ steps.detect-cache-configuration.outputs.key }}
         restore-keys: ${{ steps.detect-cache-configuration.outputs.restore-keys }}
 
+    - name: Create temporary global.json with desired .NET SDK version
```
Piedone: Use American Titles like the rest here.

Piedone: Add a comment about why we're doing this.

.github/actions/build-dotnet/action.yml
```diff
@@ -137,11 +137,19 @@ runs:
...
+      shell: pwsh
+      working-directory: ${{ inputs.directory }}
+      run: |
+        "{ ""sdk"": { ""version"": ""$Env:DOTNET_SDK_TO_USE"" } }" > ./global.json
```
Piedone: Only create a file if it doesn't exist (this is going into the root of the repo clone, right?).

.github/actions/build-dotnet/action.yml
```diff
@@ -137,11 +137,19 @@ runs:
...
+
     - name: Build and Static Code Analysis
       id: build
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
+        Write-Output ("Using .NET SDK from ./global.json: $(Get-Content .\global.json)")
```
Piedone:
```suggestion
        Write-Output "Using .NET SDK from ./global.json: $(Get-Content .\global.json)"
```

Also, a line break before the file content would be nice, because it may contain a formatted JSON, not just a single line.

.github/actions/create-jira-issues-for-community-activities/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
         PR_TITLE: ${{ github.event.pull_request.title }}
       shell: pwsh
       run: |
-        $output = "${{ github.event.pull_request }}" -eq "" ? "False" : (Confirm-PullRequestTitle "$($Env:PR_TITLE)") -or $Env:GITHUB_HEAD_REF -Like "*issue*"
+        $output = "${{ github.event.pull_request }}" -eq "" ? "False" : (Confirm-PullRequestTitle "$Env:PR_TITLE") -or $Env:GITHUB_HEAD_REF -Like "*issue*"
```
Piedone, line 35: Do we actually need quotes around `$Env:PR_TITLE`, given that the variable is the sole content of that string parameter?

.github/actions/setup-dotnet/action.yml
```diff
@@ -26,7 +26,15 @@ runs:
...
     # released .NET version for wildcard versions. So, e.g. with 6.0.x specified it'll still install 6.0.402 if the VM
     # only has 6.0.401. See: https://github.com/actions/setup-dotnet/issues/284#issuecomment-1276250960.
-    - name: Set up .NET
-      uses: actions/setup-dotnet@v3.0.2
+    - name: Set up .NET SDK
+      uses: actions/setup-dotnet@v3.2.0
+      id: setup-dotnet
       with:
         dotnet-version: ${{ inputs.dotnet-version }}
+        dotnet-quality: ga
```
Piedone, line 34: Isn't this the default?

.github/actions/setup-dotnet/action.yml
```diff
@@ -26,7 +26,15 @@ runs:
...
+
+    - name: Save .NET SDK version into Environment Variable
```
Piedone:
```suggestion
    - name: Save .NET SDK Version Into Environment Variable
```

.github/workflows/build-and-test-dotnet.yml
```diff
@@ -158,7 +158,7 @@ jobs:
           token: ${{ secrets.CHECKOUT_TOKEN }}
 
       - name: Set up .NET
-        uses: Lombiq/GitHub-Actions/.github/actions/setup-dotnet@dev
```
Piedone: Also update the default .NET version in this workflow.

Piedone (changes requested):

.github/actions/build-dotnet/action.yml
0liver: That's correct. Well, for the NuGet test solution, it's the _NuGetTest_ folder.

.github/actions/create-jira-issues-for-community-activities/action.yml
0liver: In case it contains spaces, they might be necessary.

I just removed the superfluous extra `$()` wrapper because this was the only place in LGHA where we had that.

.github/actions/setup-dotnet/action.yml
0liver: No. See https://github.com/actions/setup-dotnet/#supported-version-syntax:

image

.github/workflows/build-and-test-dotnet.yml
0liver: Thanks. I reverted the change in this file accidentally.

.github/actions/build-dotnet/action.yml
Piedone: Yeah, that's OK.

.github/actions/build-dotnet/action.yml
Piedone: This part is missing:

> Also, a line break before the file content would be nice, because it may contain a formatted JSON, not just a single line.

Piedone (changes requested):
See https://github.com/Lombiq/GitHub-Actions/pull/237#discussion_r1221739780

.github/actions/build-dotnet/action.yml
0liver: My bad.

.github/actions/build-dotnet/action.yml
```diff
@@ -137,22 +137,29 @@ runs:
...
       # We do this to force `dotnet build` to use that exact version and avoid breaking builds by silent upgrades.
       shell: pwsh
       working-directory: ${{ inputs.directory }}
       run: |
-        if (-not (Test-Path -Path .\global.json -PathType Leaf))
-        {
-          "{ ""sdk"": { ""version"": ""$Env:DOTNET_SDK_TO_USE"" } }" > .\global.json
+        $globalJsonExists = Test-Path -Path .\global.json -PathType Leaf
+        # Read from file or use empty object.
+        $globalJsonContent = if ($globalJsonExists) { Get-Content .\global.json -Raw } else { '{}' }
```
Piedone: You can use a ternary instead.

Piedone (changes requested):

.github/actions/build-dotnet/action.yml
Piedone: Nevermind, did this.

0liver: Thanks!

PR #243, opened 2023-07-02
OSOE-517: Automate managing issue branch references in workflow files

### LINKED ISSUE #168: Automate managing issue branch references in workflow files (OSOE-517)
When you're working on something related to GHA, you'll have to add issue branch references in workflow files when calling reusable workflows or actions, e.g. instead of `@dev` you use `@issue/OSOE-123`. It's painful to have to maintain these: You can forget to add them during development, and you can forget to remove them after a review. So, let's find a better way to do this. Ideally, we wouldn't need to think about managing these references for our own GHA code at all.

- @DAud-IcI mentioned that `validate-pull-request` could check for issue branch references on pushes to the default branch and fail if founds one. While this doesn't solve the management issue, it at least will tell you if you forgot to switch back to `@dev`.
- Automatically managing these would necessitate commits to the repos (including superprojects and submodules). This looks like an issue with submodules.
- A pre-commit Git hook could solve this locally (replace references to ones matching the current branch) but that would need everyone to configure it.
- Handle this in our `checkout` action: If the branch is not the default, replace all `@<default branch>` references with `@<current branch>` in all YML files. This might be a performance impact, we need to check. Also note that while this would cover projects like Lombiq GitHub Actions well (since when you open an issue branch in there, it's correct to point all internal references to that branch), it gets trickier with something like OSOCE: Should the workflows there, that call reusable workflows from LGHA, point to an issue branch? Those will only work if you have a corresponding branch in LGHA, so perhaps not (but handling this only in LGHA may be good enough as a start).
- There doesn't seem to be a built-in way to do this, or a third-party solution.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-517)

davidpuplava:
[OSOE-517](https://lombiq.atlassian.net/browse/OSOE-517)
Fixes #168 

[OSOE-517]: https://lombiq.atlassian.net/browse/OSOE-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/validate-gha-refs.yml
```diff
@@ -0,0 +1,68 @@
+name: Validate GHA References
+
+on:
+  workflow_call:
+    inputs:
+      path-patterns:
+        required: false
+        type: string
+        default: ".github"
+        description: A comma separated list of path patterns (relative to repository root) in which to check for GHA files
```
DemeSzabolcs: Put periods at the end of the sentences `.`. Also in other places and PS comments too.

.github/workflows/validate-gha-refs.yml
```diff
@@ -0,0 +1,68 @@
+name: Validate GHA References
```
DemeSzabolcs: Rather then creating a separate workflow for this, I think it should be an action, called in the [validate-this-pull-request](https://github.com/Lombiq/GitHub-Actions/blob/bf8e26ed06752c73e862f573ad22d5d15b82c9a5/.github/workflows/validate-this-pull-request.yml#L7) workflow as mentioned here: https://github.com/Lombiq/GitHub-Actions/issues/168#:~:text=%40DAud%2DIcI%20mentioned%20that%20validate%2Dpull%2Drequest%20could%20check%20for%20issue%20branch%20references%20on%20pushes%20to%20the%20default%20branch%20and%20fail%20if%20founds%20one.%20While%20this%20doesn%27t%20solve%20the%20management%20issue%2C%20it%20at%20least%20will%20tell%20you%20if%20you%20forgot%20to%20switch%20back%20to%20%40dev.

> @DAud-IcI mentioned that validate-pull-request could check for issue branch references on pushes to the default branch and fail if founds one. While this doesn't solve the management issue, it at least will tell you if you forgot to switch back to `@dev`.

It could be in the normal `validate-pull-request` workflow, but since there is an agreement, that for now we should only check Lombiq.GitHub.Actions files, then it's okay to have it in the `validate-this-pull-request` workflow.

.github/workflows/validate-gha-refs.yml
```diff
@@ -0,0 +1,68 @@
...
+    name: Check Workflows/Actions
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+
+      - name: Find Ref Mismatches
+        shell: pwsh
+        run: |
+          $repoPatterns = "${{ inputs.repo-patterns }}".Split(",")
```
DemeSzabolcs: You can use PowerShell arrays instead of separating them by commas, then splittitng them. Like here: https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/reset-azure-environment.yml#L97-L101

.github/workflows/validate-gha-refs.yml
```diff
@@ -0,0 +1,68 @@
...
+          for($i = 0; $i -lt $repoPatterns.Count; $i++)  { $repoPatterns[$i] = "uses:\s*" + $repoPatterns[$i]; }
```
DemeSzabolcs: I think this can be a `foreach` loop. Also formatting `for($i...` => `for ($i...` Same below.

.github/workflows/validate-gha-refs.yml
```diff
@@ -0,0 +1,68 @@
...
+
+          $mismatches = Get-ChildItem -Path ${{ inputs.path-patterns }} -Include ${{ inputs.include-patterns }} -Force -Recurse | 
+          Select-String -Pattern $repoPatterns | 
+          Select-String -Pattern '@${{ inputs.expected-ref }}$' -NotMatch
+          
+          if ($mismatches.Count -gt 0)
+          {
+            "Please update the following lines to match the expected GitHub Actions reference: '${{ inputs.expected-ref }}'" >> $env:GITHUB_STEP_SUMMARY
+
+            for($i = 0; $i -lt $mismatches.Count; $i++)
```
DemeSzabolcs: Shouldn't this also replace the references, not just warn about them? Or did we scratch that? Also, did we scratch the idea to replacing every `dev` references on a PR?

> If the branch is not the default, replace all `@<default branch>` references with `@<current branch>` in all YML files.

.github/workflows/validate-this-gha-refs.yml
```diff
@@ -0,0 +1,10 @@
+name: Validate GHA Refs
+on:
+  push:
+    branches:
+      - dev
+
+jobs:
+  validate-gha-references:
+    name: Validate GHA Refs
+    uses: Lombiq/GitHub-Actions/.github/workflows/validate-gha-refs.yml@main
```
DemeSzabolcs: I think this is `@dev` not `@main`.

DemeSzabolcs (changes requested):

.github/workflows/validate-gha-refs.yml
davidpuplava: Added periods at the end of descriptions and PS comments.

.github/workflows/validate-gha-refs.yml
davidpuplava: I moved the logic into an action, and moved the PowerShell code into it's own .ps1 file to be consistent with how other actions using PowerShell.

I put the logic in it's own workflow based on a comment from @BenedekFarkas to run this check only on pushes to the dev branch.
 >I think this is a good compromise in terms of complexity vs. how much it achieves. I'd rather put this in a separate workflow, because you're not really validating the pull request anymore. Although it's important that whoever merged it (or contributed to the PR) is notified in case of a failure, but that's up to personal settings too.

I temporarily added the `issue/OSOE-517` branch to the push trigger to demonstrate what the output summary would look like. I avoided adding `pull_request` as a trigger for now because so many ref mismatches are found. I think this is still a painful developer experience.

To solve this, I think I can add a step to the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow to get a list of changed/added files that are part of the pull request, and only consider those when checking GHA refs. 

This way the workflow only fails on files that are important to check. For example, if a PR is opened to change the README.md file, it is not important to check the refs of all the GHA files because the PR only concerns the readme file.

Before I implement that, though, I want to get your feedback to make sure that is desired. 

I attempted to implement it but it is more work than I expected. But I think it sort of addresses the management issue.

.github/workflows/validate-gha-refs.yml
davidpuplava: Thank you! I did not know you could pass around array data like that. I went ahead and updated the other inputs to use PowerShell arrays also.

.github/workflows/validate-gha-refs.yml
davidpuplava: I changed the loop below to use `foreach` but for this loop that prefixes the `uses:\s*`, I ended up using the following code because `foreach` didn't seem to actually modify the elements in the array in place.
```
$CalledRepoBaseIncludeList.ForEach({ "uses:\s*" + $_ })
```
The following post lead me to use the `ForEach` method on the array object itself. https://stackoverflow.com/questions/41412470/appending-a-string-to-each-item-of-an-array

.github/workflows/validate-gha-refs.yml
davidpuplava: We scratched having the workflow update all the references automatically (for now) because of the following I think
>Automatically managing these would necessitate commits to the repos (including superprojects and submodules). This looks like an issue with submodules.

And if I can add a step to the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow to consider only those files that have changed in the commit, then I think we can avoid unnecessarily changing GHA refs.

.github/workflows/validate-this-gha-refs.yml
davidpuplava: Good catch, thank you. For now, I had to change to `issue/OSOE-517` so that the [validate-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow could run on the push to the `issue/OSOE-517` branch.

.github/workflows/validate-gha-refs.yml
DemeSzabolcs: I see, okay thank you!

.github/workflows/validate-gha-refs.yml
DemeSzabolcs: > I moved the logic into an action, and moved the PowerShell code into it's own .ps1 file to be consistent with how other actions using PowerShell.

Yes that's nice!

> I put the logic in it's own workflow based on a comment from @BenedekFarkas

Ah, sorry I somehow skipped that. Then okay!

> To solve this, I think I can add a step to the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow to get a list of changed/added files that are part of the pull request, and only consider those when checking GHA refs.

How this would work exactly? I guess for `dev` it would be the same => checking everything.

And for a PR, if I would modify let's say the [Login To Azure action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/login-to-azure/action.yml) on `issue/xxx-123` branch, then it would check if where that action is used, it's using the `issue/xxx-123` branch, and only warn about those. Is that correct?

davidpuplava: >How this would work exactly? I guess for dev it would be the same => checking everything.

Yes, for `dev` it's the easier case because we can check everything.

>And for a PR, if I would modify let's say the [Login To Azure action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/login-to-azure/action.yml) on issue/xxx-123 branch, then it would check if where that action is used, it's using the issue/xxx-123 branch, and only warn about those. Is that correct?

Yes, that is correct. If the `Login to Azure` action is modified in any way, then any GHA file that calls it with `uses: Lombiq/GitHub-Actions/.github/actions/login-to-azure@...` will need to be checked for using `@issue/xxx-123`. 

I just realized that this will be tricky with actions because any of the files under the action's folder might change, not just the `.yaml` file.

Maybe the straightforward solution is to add a new `Update-GHA-Refs-to-Branch.ps1` script to the [Scripts](https://github.com/Lombiq/GitHub-Actions/tree/dev/Scripts) folder that can update all the `uses: Lombiq/GitHub-Actions/...@dev` refs to `uses: Lombiq/GitHub-Actions/...@issue/xxx-123`. Then the management issue is solved by a single PowerShell script the developer can run manually before they push their pull request, or after they get the warnings from the failed workflow check. 

What do you think? As someone who reviews PRs and then has to merge them, does adding that `Update-GHA-Refs-to-Branch.ps1` script make more sense?

.github/workflows/validate-this-gha-refs.yml
BenedekFarkas: Not sure if you'll need that, but just wanted to mention that you can grab the default branch of the current repo with `${{ github.event.repository.default_branch }}`.

davidpuplava: Thanks @BenedekFarkas - Yes, I can use that instead of hardcoding `dev`

.github/workflows/validate-gha-refs.yml
DemeSzabolcs: The scripts sounds good to me!

.github/actions/get-changed-gha-items/Get-GHA-Items-From-File-List.ps1
```diff
@@ -0,0 +1,26 @@
...
+$actions = $actionFiles.ForEach({ $PSItem.Replace("/" + $(Get-Item $PSItem).Name, "") }) | select -Unique
+
+# Filter workflow files excluding action yaml file names.
+$workflows = $FileIncludeList | Where-Object -FilterScript { 
+    (Get-Item $PSitem).BaseName -ne "action" -and 
+        ((Get-Item $PSitem).Extension -eq ".yml" -or 
+         (Get-Item $PSitem).Extension -eq ".yaml")
+}
+
+# Combine actions are workflows.
```
DemeSzabolcs: I guess this is a typo.

```suggestion
# Combine actions and workflows.
```

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+$CalledRepoBaseIncludeList = $CalledRepoBaseIncludeList.ForEach({ "uses:\s*" + $_ })
+
+$mismatchRefs = Get-ChildItem -Path $PathIncludeList -Include $FileIncludeList -Force -Recurse | 
+Select-String -Pattern $CalledRepoBaseIncludeList | 
+Select-String -Pattern $ExpectedRef -NotMatch
+
+if ($mismatchRefs.Count -gt 0)
+{
+        "These called worflows and actions do not match expected ref '$ExpectedRef'." >> $env:GITHUB_STEP_SUMMARY
```
DemeSzabolcs:
```suggestion
        "These called workflows and actions do not match expected ref '$ExpectedRef'." >> $env:GITHUB_STEP_SUMMARY
```

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,44 @@
...
+    [String[]] $CalledRepoBaseIncludeList, 
+    [String[]] $PathIncludeList,
+    [String[]] $FileIncludeList,
+    [String] $ExpectedRef,
+    [String] $GitHubRepository,
+    [String] $GitHubRefName
+)
+
+if ($CalledRepoBaseIncludeList.Count -eq 0) {
+    exit 0 # Nothing to check because array is empty.
```
DemeSzabolcs, line 13: Hmm. Maybe an informative log message would be good here.

DemeSzabolcs (changes requested):

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
BenedekFarkas: Szabolcs, can you please check the spell-checking configuration? This should've been caught by the action.

DemeSzabolcs: It ran, and I think it saw this file. So good question what happened. https://github.com/Lombiq/GitHub-Actions/actions/runs/5695397460/job/15438471092?pr=243#step:3:82

.github/actions/get-changed-gha-items/Get-GHA-Items-From-File-List.ps1
davidpuplava: Good catch, thank you, fixed!

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
davidpuplava: I fixed the spelling error. I think files with a .ps1 extension are excluded from spelling check based on [this excludes file](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/spelling/excludes.txt#L22) in the spelling action. 

Do note that I removed that line from the excludes.txt file in this branch to demonstrate how changing any file under and action can trigger the verify-gha-refs workflow to find it.

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
davidpuplava: Agreed. I added a warning message; here is a screenshot from my test repository of what the summary would look like:
image

Also note, that I made an additional change to skip the `verify-gha-refs` step altogether when no GHA items (meaning actions or workflows) are part of the PR. Specifically, I added [this line](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs-pr.yml#L39) to make the step conditional.

.github/actions/verify-gha-refs/Check-Called-GHA-Refs.ps1
DemeSzabolcs: Really good catch, thank you!

.github/workflows/validate-this-gha-refs-pr.yml
```diff
@@ -0,0 +1,76 @@
...
+  check-pr-reviews:
+    runs-on: ubuntu-22.04
+    permissions:
+      actions: read
+      contents: read
+      pull-requests: read
+    outputs:
+      approved-count: ${{ steps.check-pr-reviews.outputs.approved-count }}
+      request-changes-count: ${{ steps.check-pr-reviews.outputs.request-changes-count }}
+      comment-count: ${{ steps.check-pr-reviews.outputs.comment-count }}
```
DemeSzabolcs: Do we need these? I think they are not used (also in the other places).

DemeSzabolcs (changes requested):

DemeSzabolcs:
@davidpuplava Also some of the workflows are failing, but not for a "valid" reason: 

- https://github.com/Lombiq/GitHub-Actions/actions/runs/5826908485/job/15801860311?pr=243
- https://github.com/Lombiq/GitHub-Actions/actions/runs/5814570916/job/15764522799?pr=243
- https://github.com/Lombiq/GitHub-Actions/actions/runs/5814571224/job/15764529141?pr=243

.github/workflows/validate-this-gha-refs-pr.yml
davidpuplava: Good catch - no, not anymore. I originally (and incorrectly) assumed that Request for Changes reviews would change to Approved reviews when accepted, and the `if:` condition was `approved > 0 && changes == 0`. But instead I needed to look at the most recent review to determine if it was approved. I'll remove these and resubmit.

davidpuplava:
>Also some of the workflows are failing, but not for a "valid" reason:

Interesting. The first two are because of the PowerShell name is case sensitive, I think. The last one, I'm not so sure. I will investigate and fix.

davidpuplava:
Interestingly enough, here is an example of this [workflow run](https://github.com/Lombiq/GitHub-Actions/actions/runs/5828750349?pr=243) is an example of this feature working. I removed the `\..ps1$` entry from [excludes.txt](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/actions/spelling/excludes.txt) for the [spelling action](https://github.com/Lombiq/GitHub-Actions/tree/issue/OSOE-517/.github/actions/spelling) to catch any typos in my PowerShell scripts.. 

I also changed the description (`need to` ==> `must`) in a comment of the [validate-submodule-pull-request](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-submodule-pull-request.yml) workflow.

Glad that worked. I can also remove those changes to spelling and validation-submodule-pull-request so as not to mix it with this change.

.github/workflows/validate-this-gha-refs-pr.yml
```diff
@@ -0,0 +1,73 @@
...
+on:
+  pull_request:
+  
+jobs:
+  check-pr-reviews:
+    runs-on: ubuntu-22.04
+    permissions:
+      actions: read
+      contents: read
+      pull-requests: read
```
DemeSzabolcs: Do these permissions not granted by default?

davidpuplava: Sorry about that - I accidentally copied these over from the private repo I was testing in. I've pushed a new commit to remove them.

DemeSzabolcs:
Okay, now it looks good to me. I will test this by approving it.

DemeSzabolcs (approved):

DemeSzabolcs:
@davidpuplava  I approved the PR and `Validate GHA Refs` failed which is good since, it should warn us to use `dev` GHA actions and workflows. However it failed with a different problem: https://github.com/Lombiq/GitHub-Actions/actions/runs/5861916530/job/15892860607?pr=243#step:4:25

FYI I will be out of office from 08.17. to 08.24. Until then, I will try to review as soon and frequently as possible.

davidpuplava:
>FYI I will be out of office from 08.17. to 08.24. Until then, I will try to review as soon and frequently as possible.

Ok, I'll will try and get a fix as soon as possible. This error is unexpected.

davidpuplava:
> However it failed with a different problem: https://github.com/Lombiq/GitHub-Actions/actions/runs/5861916530/job/15892860607?pr=243#step:4:25

Ok, this was a simple copy/paste mistake on my part. The workflow file [incorrectly had conditional an if conditional](https://github.com/Lombiq/GitHub-Actions/actions/runs/5861916530/workflow?pr=243#L14) on the Checkout Repository action. This was supposed to only be under the next action which checks GHA refs on push events.

DemeSzabolcs (approved):

DemeSzabolcs:
@davidpuplava Approved it again, now looks good. One question, can't we DRY these two somehow?
image

Since they both call the same action: `actions/verify-gha-refs`. And in this case it's running twice.

davidpuplava:
> One question, can't we DRY these two somehow?

Short answer, yes. I think I can combine the two workflows and make the PR specific steps conditional on `if: github.event_name == 'pull_request'`. There will be two separate, conditional Checkout steps, only because Lombiq's [Checkout Action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/checkout/action.yml) doesn't support the `fetch-depth` (although I could add it). The key will be to make sure that the [Determine Expected Ref for GHA Files](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs-pr.yml#L50) step sets all the right inputs for each scenario (push and pull_request).

BenedekFarkas:
> only because Lombiq's [Checkout Action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/checkout/action.yml) doesn't support the fetch-depth (although I could add it)

Adding it would be nice, just make sure the default set for this parameter in our action is the same as the GH-owned one.

davidpuplava:
Hi @DemeSzabolcs

>  I think I can combine the two workflows

I've combined the two workflows into a single workflow file and put conditionals where appropriate. On my first push, the [following run](https://github.com/Lombiq/GitHub-Actions/actions/runs/6105588177) showed all the refs I needed to change back to `@dev` since the PR already had an approval on it - which is expected.

But notice that [this latest run](https://github.com/Lombiq/GitHub-Actions/actions/runs/6105618274) failed checks. This is because the files don't yet exist in the `dev` branch. I don't know if there is a way around this because we want the `@ref` to be the target branch of the merge after an approval is given. But for any new actions or workflows that don't yet exist in the target, the workflow run will fail because it can't find that workflow or action in `dev`. 

I'll think about it some more, but I don't know that there is anything we can do about this. I think this may only be an issue for GHA items that are used by the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow itself.

.github/workflows/validate-this-gha-refs.yml
```diff
@@ -3,22 +3,97 @@ on:
   push:
     branches:
       - dev
+  pull_request:
   pull_request_review:
     types: [submitted]
-
+
```
DemeSzabolcs: There is an extra space here.

.github/workflows/validate-this-gha-refs.yml
```diff
@@ -3,22 +3,97 @@ on:
...
-
+  
 jobs:
+  check-pr-reviews:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-22.04
+    permissions:
+      actions: read
+      contents: read
+      pull-requests: read
```
DemeSzabolcs: I think these permissions are again not needed.

.github/workflows/validate-this-gha-refs.yml
```diff
@@ -3,22 +3,97 @@ on:
...
+    steps:
+      - name: Check PR Reviews
+        id: check-pr-reviews
+        uses: Lombiq/GitHub-Actions/.github/actions/check-pull-request-reviews@dev
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   validate-gha-refs:
+    needs: check-pr-reviews
+    if: always()
```
DemeSzabolcs: Why do we need `always()` here? According to the documentation this will _"Causes the step to always execute, and returns true, even when canceled."_ But is there any scenario when this would be cancelled?

DemeSzabolcs:
Hello @davidpuplava,

> I'll think about it some more, but I don't know that there is anything we can do about this. I think this may only be an issue for GHA items that are used by the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow itself.

Yes, I think it's just a problem for this PR in particular because it wants to use the action. But shouldn't be a problem for anywhere else, so we can ignore it. Though I think it's a good idea to test everything again after the PR is merged (in another test PR). I will do a review.

DemeSzabolcs (changes requested):

BenedekFarkas:
> Hello @davidpuplava,
> 
> > I'll think about it some more, but I don't know that there is anything we can do about this. I think this may only be an issue for GHA items that are used by the [validate-this-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml) workflow itself.
> 
> Yes, I think it's just a problem for this PR in particular because it wants to use the action. But shouldn't be a problem for anywhere else, so we can ignore it. Though I think it's a good idea to test everything again after the PR is merged (in another test PR). I will do a review.

Wouldn't that be also a problem for a PR later on that adds a new action?

DemeSzabolcs:
> Wouldn't that be also a problem for a PR later on that adds a new action?

Well thinking about that, it actually could be a problem, but only if the action, runs in the PR where it's added. So for example, if we add a new type of "deploy" action, it won't fail, since it won't run in the PR. But something like spelling, that runs every time would fail (of course spelling is a wrong example since it's already on `dev`). 

I don't think that's a frequent case though. I don't know about any simple solution to this problem right now.

.github/workflows/validate-this-gha-refs.yml
davidpuplava: Thank you - I believe I fixed this but I also make a larger code change to the whole file.

.github/workflows/validate-this-gha-refs.yml
davidpuplava: Yep, thank you. Should be removed now.

.github/workflows/validate-this-gha-refs.yml
davidpuplava: Good question. The `check-pr-reviews` job was conditional to only run on `pull_request` events so that job would be skipped on `push` or `pull_request_review` events, and consequently, the `validate-gha-refs` job would also be skipped even thought there were steps that needed to be run for the non-pull_request scenario.

I realized this was confusing. So I refactored the code to pull the `check-pr-review` job's [one and only step](https://github.com/Lombiq/GitHub-Actions/blob/1f1e6e2fd79fb06ae97963a039a7ddbfc3dfbe19/.github/workflows/validate-this-gha-refs.yml#L21-L25) into the `validate-gha-refs` job itself [here](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/workflows/validate-this-gha-refs.yml#L62-L67).

And all the steps are now conditional based on the event that triggered the workflow run.

davidpuplava:
> I don't think that's a frequent case though. I don't know about any simple solution to this problem right now.

So I may have come up with a solution. Logically speaking, any `added` files in a PR will not exist in the target branch because they are added as part of the PR. So the only `@ref` that exists is going to be the issue branch such as `@issue/OSOE-517`. 

The problem arises when `validate-gha-refs` runs after PR approval. This happens in the [Verify GHA Items Match Expected Ref (Approved PR)](https://github.com/Lombiq/GitHub-Actions/blob/1f1e6e2fd79fb06ae97963a039a7ddbfc3dfbe19/.github/workflows/validate-this-gha-refs.yml#L91-L95) step which calls the [verify-gha-refs](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-517/.github/actions/verify-gha-refs/action.yml) action. By default, the action checks all called GHA items and throws an error when any do not match the target branch (e.g. `@dev`) . For added files, this will fail like it did in [this workflow run](https://github.com/Lombiq/GitHub-Actions/actions/runs/6105618274).

The solution I came up with was to pass the list of files that changed (similar to how the pull request check works) but exclude the `added` files. New actions that called as part of the PR will point to the issue ref (e.g.  `@issue/OSOE-517`) and allow the Approved PR workflow run to pass checks.

Finally, when the PR is merged, the `push` event will trigger another run of the `verify-gha-refs` but use the defaults to ensure that all called GHA items are pointing correctly to the `@dev` branch.

davidpuplava:
> > only because Lombiq's [Checkout Action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/checkout/action.yml) doesn't support the fetch-depth (although I could add it)
> 
> Adding it would be nice, just make sure the default set for this parameter in our action is the same as the GH-owned one.

I agree @BenedekFarkas - I started to make this change by copying the input and default value from the [GH action](https://github.com/actions/checkout/blob/v3.1.0/action.yml#L56-L58) but ran into a static analysis error (or a least a warning) that says it needs to be a string.

image

I can change it to a string easily enough but wasn't sure if it would cause issues for all the other workflows using LGHA checkout action.

BenedekFarkas:
> I can change it to a string easily enough but wasn't sure if it would cause issues for all the other workflows using LGHA checkout action.

That's OK, we do that already in other actions too. I don't remember the exact reason why it has to be like that (workflows do accept integer parameters), but it might be just because actions don't declare the type of parameters.

DemeSzabolcs (approved):

DemeSzabolcs:
@davidpuplava  `get-changed-files-from-git-diff` fails (since it's not in `dev` yet), and that's the one that checks the changed files. We can ignore this one I think.

DemeSzabolcs:
@davidpuplava I will merge this, but it would be good to actually test the `get-changed-files-from-git-diff` when it's on the `dev` branch in another PR.

PR #244, opened 2023-07-13
CARE-368: Use pre-release version to debug "The operation was canceled" error

0liver:
[CARE-368](https://lombiq.atlassian.net/browse/CARE-368)


0liver:
Fixed the underlying problem in CARE. Not needed anymore.

Issue #22, opened 2022-06-30
Remove setup-edge-linux once Edge is available on the Ubuntu image


Piedone:
See https://github.com/actions/virtual-environments/issues/5845

PR #247, opened 2023-07-26
OSOE-655: Only run Post Pull Request Checks Automation if the current workflow is the most recent one in Lombiq.GitHub.Actions

### LINKED ISSUE #241: Only run Post Pull Request Checks Automation if the current workflow is the most recent one (OSOE-655)
Post Pull Request Checks Automation will run, and thus e.g. transition Jira issues or merge the PR, in any case. This is an issue because in the likes of the [OSOCE build workflow](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/blob/3f8e9b85bc1905103c64e8227f09fd0a2a5fb638/.github/workflows/build-and-test.yml#L66), it'll also run if the current workflow run is not the most recent. E.g. if you add the "done-jira-issue-if-checks-succeed" label, it'll transition the Jira issue even if there is a more recent commit and that has a failing workflow run. So, let's add a check to it that'll only run the workflow if the current workflow run is the most recent one.

Note that the current workflow run being the most recent one isn't necessarily the same as the most recent commit for the current PR, since that can contain `[skip ci]` too.

Let's have this as an optional behavior of the Post Pull Request Checks Automation workflow, but enabled by default.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-655)

Psichorex:
[OSOE-655](https://lombiq.atlassian.net/browse/OSOE-655)
Fixes #241

[OSOE-655]: https://lombiq.atlassian.net/browse/OSOE-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -45,9 +48,26 @@ on:
...
 
 jobs:
+  get_latest_workflow:
+    name: Get the latest workflow run number
+    runs-on: ubuntu-latest
+    outputs:
+      latest_run_number: ${{ steps.get_latest_workflow_run_number.outputs.latest_run_number }}
+    steps:
+      - name: Get the latest workflow run number
+        id: get_latest_workflow_run_number
```
Piedone: Follow naming conventions for the job ID, job name, output name, step ID.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -1,5 +1,8 @@
 name: Post Pull Request Checks Automation
 
+env:
+  RUN_ONLY_LATEST_WORKFLOW: 'true'
```
Piedone: This doesn't need to be an env var and can be an input instead.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -45,9 +48,26 @@ on:
...
+      latest_run_number: ${{ steps.get_latest_workflow_run_number.outputs.latest_run_number }}
+    steps:
+      - name: Get the latest workflow run number
+        id: get_latest_workflow_run_number
+        run: |
+          LATEST_RUN_NUMBER=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+                             -H "Accept: application/vnd.github.v3+json" \
+                             "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=push" \
+                             | jq '.workflow_runs[0].run_number')
+          echo "::set-output name=latest_run_number::$LATEST_RUN_NUMBER"
```
Piedone: This doesn't need to be a separate job, can be a step in `post-pull-request-checks-automation`.

Piedone: Use `Set-GitHubOutput` instead.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -45,9 +48,26 @@ on:
...
+  get_latest_workflow:
+    name: Get the latest workflow run number
+    runs-on: ubuntu-latest
+    outputs:
+      latest_run_number: ${{ steps.get_latest_workflow_run_number.outputs.latest_run_number }}
+    steps:
+      - name: Get the latest workflow run number
+        id: get_latest_workflow_run_number
+        run: |
+          LATEST_RUN_NUMBER=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
```
Piedone: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=When%20needing%20to%20interact%20with%20the%20GitHub%20web%20API

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -45,9 +48,26 @@ on:
...
+                             -H "Accept: application/vnd.github.v3+json" \
+                             "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=push" \
+                             | jq '.workflow_runs[0].run_number')
```
Piedone: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=Prefer%20to%20use%20PowerShell%207%2B%20scripts

Piedone (changes requested):

.github/workflows/post-pull-request-checks-automation.yml
Psichorex: Do you mean the whole job could be a part ? Or just the steps part?

Piedone: Not part, a step. You can take the only step in this job, and put it as the first step in the other job, and then add the condition check to the subsequent steps.

Psichorex: What I meant is that currently this `get_latest_workflow` job becomes redundant if I remove the step inside of it. 
So what I meant is that remove this whole job and put its step in the other job.
Is this what you meant?

Piedone: Yes.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
         type: number
         default: 1
         description: Configuration for the timeout-minutes parameter of the workflow.
+      run-only-latest-workflow:
+        type: string
+        description: 'Run only if the current workflow is the most recent workflow.'
```
Piedone:
```suggestion
        description: Run only if the current workflow is the most recent workflow.
```

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
...
+        default: 'true'
```
Piedone:
```suggestion
        default: true
```

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
...
 
 jobs:
   post-pull-request-checks-automation:
+    outputs:
+      latest-run-number: ${{ steps.get-latest-workflow-run-number.outputs.latest_run_number }}
```
Piedone: Why do we need this output? This is only if you need the job itself to have outputs, not needed for the output of steps.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
...
     runs-on: ubuntu-22.04
     timeout-minutes: ${{ inputs.timeout-minutes }}
     steps:
+      - name: Checkout
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
Piedone: This shouldn't be needed.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
...
+
+      - name: Get Latest Workflow Run Number
+        id: get-latest-workflow-run-number
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          LATEST_RUN_NUMBER=$(gh run list --limit 1 --json databaseId | jq -r '.[0].databaseId')
+          echo "latest_run_number=$LATEST_RUN_NUMBER" >> $GITHUB_OUTPUT
```
Piedone: Again, use `Set-GitHubOutput` instead. And still follow naming conventions for outputs.

Piedone: Still https://lombiq.atlassian.net/wiki/spaces/DEV/pages/2545614855/Writing+GitHub+Actions+workflows+and+actions#:~:text=Prefer%20to%20use%20PowerShell%207%2B%20scripts

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,49 @@ on:
...
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get Latest Workflow Run Number
+        id: get-latest-workflow-run-number
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          LATEST_RUN_NUMBER=$(gh run list --limit 1 --json databaseId | jq -r '.[0].databaseId')
```
Piedone: This only needs to run in `run-only-latest-workflow` is `true`. So when that's not the case, cancel running the other steps. For this, so the name conveys the meaning better, rename the output to something like `should-run`, make it a bool and thus do the `run_id` check here, and only check in later steps on that.

Piedone (changes requested):

.github/workflows/post-pull-request-checks-automation.yml
Psichorex: Without this it doesn't let me interact with GH CLI and GH run list doesn't work.

.github/workflows/post-pull-request-checks-automation.yml
Psichorex: My problem with that is `Set-GithubOutput` is not being recognized by the cmdlet at least that is what GH is indicating.
Problem something is missing a prerequisite for example but I didn't find any in the similar files.
Do you know anything that should be included for `Set-GithubOutput` so that it can be used?

.github/workflows/post-pull-request-checks-automation.yml
Piedone: What issue are you seeing, specifically?

.github/workflows/post-pull-request-checks-automation.yml
Piedone: Yes, see the "Setup Scripts" steps in various actions (note that you need a composite action, not reusable workflow, for that to work).

.github/workflows/post-pull-request-checks-automation.yml
Psichorex: image

Possible that now we use pwsh this might work. It was not working with a bash script that is for sure.
But even now with pwsh it is still alerting that the repo is messy.

Piedone: This is the same as: https://github.com/Lombiq/GitHub-Actions/pull/247#discussion_r1278604849

Psichorex: The above line "failed to determine base repo: failed to run git: fatal: not a git repository"
WIll probably occure due to the `gh run list` command without a checkout.

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
+name: Check current workflow is the latest workflow
```
Piedone: Use naming conventions here too.

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+description: Checks whether the current workflow is the most recent workflow.
```
Piedone: Similar to other internally uses actions, add "Intentionally not documented in Actions.md since it's only meant for internal use."

Piedone: ...for the current pull request?

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
+        (Resolve-Path "${{ github.action_path }}/../../../Scripts").Path >> $Env:GITHUB_PATH
+      
+    - name: Check Current Workflow Equals Latest
+      shell: pwsh
+      id: check-current-workflow-equals-latest
+      run: |
+        $latestWorkflowRunNumber=$(gh run list --limit 1 --json databaseId | jq -r '.[0].databaseId')
```
Piedone: Instead of the Linux-only `jq`, use the cross-platform PowerShell `ConvertFrom-Json`.

Piedone: `gh run list` lists all workflow runs from the current repo, thus this line is fetching the most recent workflow run for every pull request too. This is not what we need. Only the workflow runs of the current PR are interesting, since other workflow runs are irrelevant to this action.

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+        $currentEqualsLatest= ($latestWorkflowRunNumber -eq ${{ inputs.current-workflow-id }})
```
Piedone:
```suggestion
        $currentEqualsLatest = ($latestWorkflowRunNumber -eq ${{ inputs.current-workflow-id }})
```

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+        Set-GitHubOutput -Key 'should-run' -Value $currentEqualsLatest
```
Piedone:
```suggestion
        Set-GitHubOutput 'should-run' $currentEqualsLatest
```

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+
+inputs:
+  current-workflow-id:
+    description: The current workflow run's database id.
+    required: true
+    default: ""
+
+outputs:
+  should-run:
+    description: The latest workflow run's database id.
```
Piedone: That's not what the output contains.

.github/actions/check-current-workflow-is-latest/action.yml
Piedone:
```suggestion
        $latestWorkflowRunNumber = $(gh run list --limit 1 --json databaseId | jq -r '.[0].databaseId')
```

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    value: ${{ steps.check-current-workflow-equals-latest.outputs.should-run }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Scripts
+      shell: pwsh
+      run: |
+        "${{ github.action_path }}" >> $Env:GITHUB_PATH
```
Piedone: This line isn't needed since there are no scripts for this action.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -42,28 +42,22 @@ on:
...
+        uses: Lombiq/GitHub-Actions/.github/actions/check-current-workflow-is-latest@issue/OSOE-655
+        id: check-steps-should-run
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          LATEST_RUN_NUMBER=$(gh run list --limit 1 --json databaseId | jq -r '.[0].databaseId')
-          echo "latest_run_number=$LATEST_RUN_NUMBER" >> $GITHUB_OUTPUT
+        with:
+          current-workflow-id: ${{ github.run_id }}
+        if: inputs.run-only-latest-workflow == 'true'
```
Piedone: Put all `if`s a the top of the steps as usual, after `name`.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -73,7 +67,7 @@ jobs:
           $mergeToken = $Env:MERGE_TOKEN ? $Env:MERGE_TOKEN : "${{ secrets.GITHUB_TOKEN }}"
           "MERGE_TOKEN=$mergeToken" >> $Env:GITHUB_ENV
         if: |
```
Piedone: A `|` shouldn't be needed for `if`s.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,26 +40,43 @@ on:
...
+
       - name: Set Merge Token
         shell: pwsh
         env:
           MERGE_TOKEN: ${{ secrets.MERGE_TOKEN }}
         run: |
           $mergeToken = $Env:MERGE_TOKEN ? $Env:MERGE_TOKEN : "${{ secrets.GITHUB_TOKEN }}"
           "MERGE_TOKEN=$mergeToken" >> $Env:GITHUB_ENV
+        if: |
+          (inputs.run-only-latest-workflow == 'false') || (steps.check-steps-should-run.outputs.should-run)
```
Piedone: Is the `inputs.run-only-latest-workflow == 'false'` part needed here? Since `steps.check-steps-should-run.outputs.should-run` won't be set anyway if that step hasn't run. Also, you need `steps.check-steps-should-run.outputs.should-run == 'true'`.

Piedone (changes requested):

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,13 +40,32 @@ on:
...
+        type: string
+        description: Run only if the current workflow is the most recent workflow.
+        default: true
 
 jobs:
   post-pull-request-checks-automation:
     runs-on: ubuntu-22.04
     timeout-minutes: ${{ inputs.timeout-minutes }}
     steps:
+      - name: Checkout
```
Psichorex: As far as I have tested things this is a must here because without this I cannot access data in the composite action that I need for it to work properly. As soon as I added checkout everything started working.

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -1,5 +1,7 @@
-name: Check current workflow is the latest workflow
-description: Checks whether the current workflow is the most recent workflow.
+name: Check Current Workflow Is The Latest Workflow
```
Piedone:
```suggestion
name: Check Current Workflow is the Latest Workflow
```

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -1,5 +1,7 @@
...
+description: >
+  Checks whether the current workflow is the most recent workflow for the pull request.
+  Intentionally not documented in Actions.md since it's only meant for internal use.
```
Piedone:
```suggestion
  Checks whether the current workflow is the most recent workflow for the pull request. Intentionally not documented in
  Actions.md since it's only meant for internal use.
```

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -9,7 +11,7 @@ inputs:
 
 outputs:
   should-run:
-    description: The latest workflow run's database id.
+    description: Whether subsequent jobs should run.
     value: ${{ steps.check-current-workflow-equals-latest.outputs.should-run }}
```
Piedone: Since the action is about checking whether the workflow run is a latest one, the output should also reflect that, in itself it doesn't have anything to do with subsequent steps running. So I'd rather call it e.g. `is-latest`, adjusting the other identifiers and the description too.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -40,13 +40,32 @@ on:
...
         description: Configuration for the timeout-minutes parameter of the workflow.
+      run-only-latest-workflow:
+        type: string
+        description: Run only if the current workflow is the most recent workflow.
+        default: true
 
 jobs:
   post-pull-request-checks-automation:
     runs-on: ubuntu-22.04
     timeout-minutes: ${{ inputs.timeout-minutes }}
```
Piedone, line 52: This is not your code just I've seen it: Please add `name: Post Pull Request Checks Automation`.

.github/workflows/post-pull-request-checks-automation.yml
Piedone: Hmm, this shouldn't be necessary unless you want to run actions from the OSOCE repo, which is not the case. And GHA will automatically download the LGHA repo's content even without a checkout step (but then again, this one here doesn't do that), see:

image

Also, there's no `secrets.CHECKOUT_TOKEN` here.

Please show what exactly happens without this checkout.

Piedone (changes requested):

.github/workflows/post-pull-request-checks-automation.yml
Psichorex: @Piedone 
image
This happens when there is no Checkout. GH commands basically don't work. This line is perfect is there is checkout.

Psichorex: This is the result with checkout:
image
You can see at `line #22` that the echo did bring back the variable's value and it didn't error out.

Piedone: Please figure out what the null array there is. Having to checkout the OSOCE repo to run this workflow that does nothing with the OSOCE repo's content doesn't make sense.

Psichorex: IF you ask me
image
The `failed to determine base repo:  failed to run git` part is the one to blame. It seems that for some reason we cannot access the GH CLI in a `composite actions` unless we are checking out in the REPO. 

By the way you mentioned `Having to checkout the OSOCE repo to run this workflow that does nothing with the OSOCE repo's content doesn't make sense.` 
Which is not 100% correct as the GH run list CLI command needs to access the workflows that are running in the `OSOCE` repository so if I am not mistaken we will need a checkout here.

Psichorex: So ultimately `github context` is passed through and can be accessed from a `composite action` without checking out but the `GH CLI` needs the repository to function.

Psichorex: I am still trying to work it around though

Piedone: Perhaps you need to pass `GITHUB_TOKEN` to the action from the workflow? Actions don't automatically have access to the exact same context as workflows, and thus they won't have this available in the environment either. And [the docs](https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows) tells this too:

> GitHub CLI is preinstalled on all GitHub-hosted runners. For each step that uses GitHub CLI, **you must set an environment variable called GITHUB_TOKEN** to a token with the required scopes.

Psichorex: image
This seems to be the solution here.
So basically unless we are checking out we need to specify in the `gh run list` command in which repository do we want to list the workflow. So `--repo ${{ github.repository }}` is a must when using `GH CLI`

Psichorex: @Piedone By the way there is a new flaw.
Which can be omitted with a bit more accurate querying of the runs.
So basically if you now put the `done-jira-issue-if-checks-succeed` to the PR and you expect this workflow to work. It will but it won't as the done-jira tag will trigger a skipped windows build which will eventually be a more recent workflow.
I am sure you want to get something be done about that.

Piedone: So this breaks how we work. This should be addressed before we can use this, yes.

Piedone: You can e.g. try to look for the most recent workflow of the same name than the current workflow, not just any workflow for the current branch.

Psichorex: Yes this is an option that you said but in some wierd scenarios the following could happen:
1)
Is this the latest `Build and Test ubuntu latest` ? If yes and it passed put the issue to done
BUT
What if there is a pending `Build and Test Windows` that will fail for some reason. We put the issue to done but there is a failing build.

I investigated the switches and there really isn't a good solution for it.
We could try to play with the `--status` switch but that isn't easy either as the Windows build will be mostly either 
`skipped|queued|in_progress` and the `status` flag will accept only 1.

It would be the best to have something like `gh run list --allbutskipped` but that is basically a `!skipped` which doesn't exist. 

Either way if we really want to count in the windows workflow it will be a bit messy.

Piedone: We don't need to care about Windows builds for now. It's extremely rare that you run it together with the Ubuntu ones, since the PR author most of time won't run them, and the PR reviewer will only run them at the end (at which point the latest Ubuntu build for the most recent commit should already be finished).

So, the logic here can be what I've written, i.e. the most recent workflow of the same name, for the current branch.

.github/workflows/post-pull-request-checks-automation.yml
```diff
@@ -65,7 +61,7 @@ jobs:
           current-workflow-id: ${{ github.run_id }}
 
       - name: Set Merge Token
-        if: steps.check-steps-should-run.outputs.should-run == 'true'
+        if: steps.check-steps-should-run.outputs.is-latest == 'true'
```
Piedone: Sorry for giving a misleading tip earlier, but these should actually check `!= 'false'`. That way, they will match if the value is `'true'` but also if it's not set because `inputs.run-only-latest-workflow` is `'false'`.

.github/actions/check-current-workflow-is-latest/action.yml
```diff
@@ -26,6 +26,6 @@ runs:
       shell: pwsh
       id: check-current-workflow-equals-latest
       run: |
-        $latestWorkflowRunDatabaseId = $(gh run list --branch ${{ github.head_ref }} --limit 1 --json databaseId | ConvertFrom-Json)[0].databaseId
+        $latestWorkflowRunDatabaseId = $(gh run list -R '${{ github.repository }}' -b '${{ github.head_ref }}' -w '${{ github.workflow }}' -L 1 --json databaseId | ConvertFrom-Json)[0].databaseId
```
Piedone: Line is too long: https://lombiq.atlassian.net/wiki/spaces/DEV/pages/43384918/PowerShell+code+conventions#Formatting.

Also, use unabbreviated parameter names so the command is easier to read.

Piedone (changes requested):

.github/actions/check-current-workflow-is-latest/action.yml
Psichorex: image
image
Splatting will just not work in my case for some reason so I will revert to making variable for every args.

Piedone: The docs I linked clearly state that you should use splatting with arrays in this case like:

```
$dockerRunSwitches = @(
    '--name', 'sql2019'
    '--env', 'ACCEPT_EULA=Y'
    '--env', 'SA_PASSWORD=Password1!'
    '--publish', '1433:1433'
    '--detach', 'mcr.microsoft.com/mssql/server:2019-latest'
)

docker run @dockerRunSwitches
```

So please use that.

Piedone (changes requested):

.github/actions/check-current-workflow-is-latest/action.yml
Psichorex: Well, I missunderstood the docs. Now it makes sense. For some reason I thought that the Hash Table version should be used in this case.

Piedone (approved):

PR #249, opened 2023-08-02
OSOE-664: handle explicit IsPackage = true and fix error quote

sarahelsaig:
[OSOE-664](https://lombiq.atlassian.net/browse/OSOE-664)

[OSOE-664]: https://lombiq.atlassian.net/browse/OSOE-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -83,8 +83,8 @@ foreach ($project in $projects)
         
         $messageType = $isRequired ? "error" : "warning"
         Write-Output ("::$messageType file=$($project.FullName)::Packing was skipped because $($project.Name) doesn't " +
-            'have a <PackageLicenseFile> property. You can avoid this check by including the " +
-            "<IsPackable>false</IsPackable> property.')
+            "have a <PackageLicenseFile> property. You can avoid this check by including the " +
+            "<IsPackable>false</IsPackable> property.")
 
         if ($isRequired) { exit 1 }
         continue
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -47,7 +47,6 @@ function Get-ProjectProperty
         Set-Content $temporaryProjectFilePath $updatedProjectFileContent -ErrorAction Stop
 
         $buildOutput = dotnet msbuild $temporaryProjectFilePath /nologo /v:minimal /p:DesignTimeBuild=true /p:BuildProjectReferences=false /t:GetPropertyValue
-        Write-Output "BUILD OUTPUT: '$buildOutput'"
 
         # Removing the temporary file.
         Remove-Item $temporaryProjectFilePath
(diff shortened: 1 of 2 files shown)


Piedone (approved):

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -47,6 +47,7 @@ function Get-ProjectProperty
         Set-Content $temporaryProjectFilePath $updatedProjectFileContent -ErrorAction Stop
 
         $buildOutput = dotnet msbuild $temporaryProjectFilePath /nologo /v:minimal /p:DesignTimeBuild=true /p:BuildProjectReferences=false /t:GetPropertyValue
+        Write-Output "BUILD OUTPUT: '$buildOutput'"
```
sarahelsaig, line 52: For whatever reason when I don't have this log here, the main package gets skipped. For example [here](https://github.com/Lombiq/Orchard-Privacy/actions/runs/5753728233). It looks like the new MSBuild based thing sometimes doesn't return anything but I don't know the reason. I assume it's just a coincidence, but it never failed to publish the packages correctly when these logs were present. At any rate if it ever happens again (I expect this to crop up when I continue doing OSOE-603 after this is resolved) it will be invaluable information, so I'd prefer to keep it for now. Is it okay if I open a new bug issue regarding this publication flakyness? It's not related to the current issue so I don't want to hold up the resolution here.

Piedone: This is some dark magic shit. Yeah, then please keep it here, with a comment on why, and indeed create an issue and link it in the comment. We'll keep that low prio though, since having the build output in the workflow output won't bother us.

sarahelsaig: Added https://github.com/Lombiq/GitHub-Actions/issues/250

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -85,6 +88,9 @@ foreach ($project in $projects)
             'have a <PackageLicenseFile> property. You can avoid this check by including the ' +
             '<IsPackable>false</IsPackable> property.')
 
+        Write-Output "isPackableProperty: '$isPackableProperty'"
+        Write-Output (Get-Content $project)
```
Piedone, line 92: Is this also needed for the magic fix or is this rather leftover debug code?

sarahelsaig: I left it in as context that can be helpful every time the warning above will be displayed. It's intentionally not part of the warning message but still available. I think we can feel free to log more during warnings and exceptions. Regular operation dumps a lot of text into the action anyway so I see no downside.

Piedone: OK, sounds good.

Piedone (approved):

PR #255, opened 2023-09-15
OSOE-688: Azure authentication with Service Principal Federated Credentials

### LINKED ISSUE #254: Azure authentication with Service Principal Federated Credentials (OSOE-688)
1. Update our [login-to-azure action](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/login-to-azure/action.yml) to use atomic parameters instead of a JSON.
2. Update our [the reusable workflows that include](https://github.com/search?q=org%3ALombiq%20%22login-to-azure%22&type=code) and the ones in our organization that call it to use the new parameter set.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-688)

BenedekFarkas:
[OSOE-688](https://lombiq.atlassian.net/browse/OSOE-688)
Fixes #254

[OSOE-688]: https://lombiq.atlassian.net/browse/OSOE-688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone, line 1: Update the docs too. Does it mean that we can actually use GitHub environments, as opposed to https://github.com/Lombiq/GitHub-Actions/blob/dev/Docs/Workflows.md#:~:text=To%20restrict%20who,a%20suitable%20substitute.)? Though having to put the Swap workflow into a different repo still applies if you're not on Enterprise.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
```diff
@@ -3,14 +3,22 @@ name: Deploy Orchard 1 to Azure App Service
...
       CHECKOUT_TOKEN:
         description: >
           The GitHub token to authenticate checkout. Pass in a GitHub personal access token if authenticated submodules
           are used.
-      AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL:
+      AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL_ID:
+        required: true
+      AZURE_APP_SERVICE_DEPLOYMENT_AZURE_TENANT_ID:
+        required: true
+      AZURE_APP_SERVICE_DEPLOYMENT_AZURE_SUBSCRIPTION_ID:
```
Piedone: Add a description for these. Also in the other workflows.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
```diff
@@ -3,14 +3,22 @@ name: Deploy Orchard 1 to Azure App Service
...
   group: AzureDeployApp
 
+permissions:
+  id-token: write
+  contents: read
+
 on:
   workflow_call:
     secrets:
       CHECKOUT_TOKEN:
```
Piedone, line 13: This is not new code but a new syntax error: "Missing property "required"." So please add it. Also in the other workflows.

.github/actions/login-to-azure/action.yml
Piedone, line 1: Won't we use our login script, then?

Piedone (changes requested):

.github/actions/login-to-azure/action.yml
BenedekFarkas: Nope.

Piedone: Why not?

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: Oops!

> Does it mean that we can actually use GitHub environments

I haven't moved on to exploring what more we can do with environments, but as far as I can tell now, they (alone) don't help us alleviate having to put the Swap workflows in a separate repo. Adding environments was out of necessity to be able to set up the federated credentials without having to define specific branches that are allowed to run.

.github/actions/login-to-azure/action.yml
BenedekFarkas: It wasn't designed for federated credentials.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: I mean yeah, having a Swap repo doesn't change without Enterprise. However, the linked note (in this form) doesn't apply now, and otherwise we can utilize Environments to display what's currently in that environment (as opposed to have to check the workflow list, for example), right?

.github/actions/login-to-azure/action.yml
Piedone: OK then.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I see, OK, well, that text is a bit confusing, because it mixes two separate things together: using Environments and restrictions due to not having an Enterprise subscription. Also, this is an open-source component, so we don't know if either of those applies to whoever is reading it. I'll try to rework it.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: Actually, do all of those secrets need to be required? There are probably cases (e.g., for an open source repo) where it's needed.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I don't think they each need a description, their names are quite self-explanatory, but I added some context and link to explain how they will be used.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: `CHECKOUT_TOKEN` is not always needed, but what's required is the `required` parameter :). This just needs `required: false`.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: Keep in mind that these workflows can be used potentially by anybody, including people with little experience in Azure (including our own colleagues or clients) who need more hand-holding. Since these secrets are part of the public interface of the workflow, they need to be the easiest-to-understand parts of the workflow. Naturally, we're not looking to include a tutorial, but explanations on the level of all the other parameters are necessary, just how we document the published APIs of our libraries.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I mean I can add that for sure, but the question is whether it's actually correct or some of them should be required. If yes, then reviewing that and correcting should be in a separate issue.

Piedone: I'm only talking about `CHECKOUT_TOKEN` and I'm certain it only needs `required: false` (what's the default anyway, but apparently it's necessary to spell it out).

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I always keep that in mind and as you can see there's a clear pointer to our login action, so you can see how it's used and the link to the azure/login repo, whose readme has a very detailed explanation of the authentication process. Previously (with the JSON secret) there was no explanation whatsoever.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: All right then!

BenedekFarkas: I've checked all the workflows and there were other secrets with this validation error, but none of them need to be required.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: One can surely dig up all the information but having it spelled out is better. You can of course link to the `azure/login` action's docs for further reference too; I'm not saying all the information for these secrets should be right here in the code but some pointers should. (Our `login-to-azure` action is for internal use BTW and we can't expect consumers of this workflow to follow up to that, and then from there to `azure/login`.)

The previous secret not having docs was a mistake that we can now correct.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: Thanks!

.github/workflows/deploy-orchard1-to-azure-app-service.yml
```diff
@@ -11,17 +11,23 @@ on:
...
           are used.
+
+      # These secrets are used for Azure authentication through Service Principal Federated Credentials with the
+      # azure/login (https://github.com/azure/login) action, which is proxied by our login-to-azure action below.
       AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL_ID:
         required: true
       AZURE_APP_SERVICE_DEPLOYMENT_AZURE_TENANT_ID:
         required: true
       AZURE_APP_SERVICE_DEPLOYMENT_AZURE_SUBSCRIPTION_ID:
         required: true
```
Piedone: Same for the others:

```suggestion
      AZURE_APP_SERVICE_DEPLOYMENT_SERVICE_PRINCIPAL_ID:
        required: true
        description: >
            The Application (client) ID of the Microsoft Entra ID Service Principal used for the OpenID Connect-based
            authentication with Azure services. Check out the azure/login action's documentation
            (https://github.com/azure/login) on the client-id parameter for details.
      AZURE_APP_SERVICE_DEPLOYMENT_AZURE_TENANT_ID:
        required: true
        description: >
            The tenant ID of the Microsoft Entra ID tenant used for the OpenID Connect-based authentication with Azure
            services. Check out the azure/login action's documentation (https://github.com/azure/login) on the tenant-id
            parameter for details.
      AZURE_APP_SERVICE_DEPLOYMENT_AZURE_SUBSCRIPTION_ID:
        required: true
        description: >
            The subscription ID of the Azure subscription the resources are under. You can look this up e.g. in the
            Azure Portal under any resource or the subscription itself.
```

.github/workflows/deploy-orchard1-to-azure-app-service.yml
Piedone: I'm adding code suggestions that I hope will make it clearer what exactly I find useful in this context.

Docs/Workflows.md
```diff
@@ -292,16 +294,20 @@ jobs:
       timeout-minutes: 60
       app-name: AppName
       resource-group-name: ResourceGroupName
+      source-slot-name: Production
+      destination-slot-name: Staging
```
Piedone, line 300: Same for Swap Azure Web App Slots.

```suggestion
      # These are also the default slot names but here's how you can configure them.
      source-slot-name: Production
      destination-slot-name: Staging
```

Piedone (changes requested):

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I guess it's a matter of where you want to draw line in terms of beginner-friendliness, IMO it's on the dumbing-down end of the spectrum, but OK.

.github/workflows/deploy-orchard1-to-azure-app-service.yml
BenedekFarkas: I can't commit this suggestion directly for some reason.

Piedone (approved):

PR #260, opened 2023-10-02
OSOE-636: Fixing check-spelling update to v0.0.22

Piedone:
[OSOE-636](https://lombiq.atlassian.net/browse/OSOE-636)

[OSOE-636]: https://lombiq.atlassian.net/browse/OSOE-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/spelling/action.yml
+++ b/.github/actions/spelling/action.yml
@@ -151,7 +151,8 @@ runs:
       id: step-check-spelling
       # When updating the check-spelling action to a new release, update the 'cspell' prefix in the "Merge configuration
       # source prefixes" step with the dictionary version tag that belongs to the commit of the release. See
-      # https://github.com/check-spelling/check-spelling/blob/v0.0.22/action.yml#L67 for current version.
+      # https://github.com/check-spelling/check-spelling/blob/v0.0.22/action.yml#L67 for current version. These
+      # dictionaries are under https://github.com/check-spelling/cspell-dicts/tree/main/dictionaries.
       uses: check-spelling/check-spelling@00c989c97749eb0cb2d256bdc55ac61b0096c6d3 # v0.0.22
       with:
         check_file_names: 1
--- a/.github/actions/spelling/action.yml
+++ b/.github/actions/spelling/action.yml
@@ -152,8 +152,7 @@ runs:
       id: step-check-spelling
       # When updating the check-spelling action to a new release, update the 'cspell' prefix in the "Merge configuration
       # source prefixes" step with the dictionary version tag that belongs to the commit of the release. See
-      # https://github.com/check-spelling/check-spelling/blob/v0.0.22/action.yml#L67 for current version. These
-      # dictionaries are under https://github.com/check-spelling/cspell-dicts/tree/main/dictionaries.
+      # https://github.com/check-spelling/check-spelling/blob/v0.0.22/action.yml#L67 for current version.
       # Also update check_extra_dictionaries below as necessary.
       uses: check-spelling/check-spelling@00c989c97749eb0cb2d256bdc55ac61b0096c6d3 # v0.0.22
       with:
@@ -169,9 +168,10 @@ runs:
         # While extra_dictionaries adds dictionaries that'll be used for spell-checking, the dictionaries under
         # check_extra_dictionaries will only be used to provide hints for unrecognized words. You'll see this under the
         # check-spelling report (PR comment and workflow run summary) along the lines of "Available dictionaries could
-        # cover words not in the dictionary". The values here are some of the default dictionaries and ours. When
-        # updating the check-spelling action to a new release, consider updating this list too with new entries from
-        # https://github.com/check-spelling/check-spelling/blob/main/action.yml.
+        # cover words not in the dictionary". The values here are some of the default dictionaries and ours. The default
+        # dictionaries are under https://github.com/check-spelling/cspell-dicts/tree/main/dictionaries.
+        # When updating the check-spelling action to a new release, consider updating this list too with new entries
+        # from https://github.com/check-spelling/check-spelling/blob/main/action.yml.
         check_extra_dictionaries: |
           cspell:aws/aws.txt
           cspell:companies/src/companies.txt
--- a/.github/actions/spelling/action.yml
+++ b/.github/actions/spelling/action.yml
@@ -183,10 +183,9 @@ runs:
           cspell:npm/dict/npm.txt
           cspell:powershell/dict/powershell.txt
           cspell:public-licenses/src/generated/public-licenses.txt
-          cspell:software-terms/src/network-os.txt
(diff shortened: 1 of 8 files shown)


.github/actions/spelling/action.yml
```diff
@@ -168,26 +168,27 @@ runs:
...
-          cspell:node/node.txt
-          cspell:npm/npm.txt
-          cspell:powershell/powershell.txt
+          cspell:node/dict/node.txt
+          cspell:npm/dict/npm.txt
+          cspell:powershell/dict/powershell.txt
           cspell:public-licenses/src/generated/public-licenses.txt
-          cspell:software-terms/src/network-os.txt
-          cspell:software-terms/src/network-protocols.txt
-          cspell:software-terms/src/software-terms.txt
```
Piedone, line 186: @BenedekFarkas we had this both here and under `default-dictionaries` above. Can you tell why this is the case? Actually, why do we need to use `check_extra_dictionaries` at all, couldn't we put everything into `default-dictionaries` (since these here in `check_extra_dictionaries` are also default dictionaries, but ones that can't be overridden)?

BenedekFarkas: Those have a very different purpose:

- `default-dictionaries` are those that we selected as general-purpose enough so that probably every or almost every consumer project will need them. It can be overridden, but if you just want to add to it, you can use `additional-dictionaries` instead. Those two will be merged in the `Merge dictionary lists` step, producing `merged-dictionary-list`, which will be passed on to the spelling action as `extra_dictionaries` (see: https://github.com/check-spelling/check-spelling/blob/main/action.yml#L68).
Dictionaries that are used for the actual spell checking should be kept to a minimum to avoid false negatives.
- `check_extra_dictionaries` will be passed on directly (see https://github.com/check-spelling/check-spelling/blob/main/action.yml#L72), but this list of dictionaries will be used to suggest dictionaries when the spellchecking report comes up with unrecognized entries. The entries that are also found `default-dictionaries` are kept here, so even if you override `default-dictionaries` (and remove something), the spell checking report can still suggest such a useful dictionary to be re-added.

Piedone: So `check_extra_dictionaries` is solely for this?

image

Since `check-spelling` is suggesting dictionaries from https://github.com/check-spelling/cspell-dicts/tree/main/dictionaries (or [the defaults](https://github.com/check-spelling/check-spelling/blob/ed14e107b2404ceec4db32161bc3cc353b123b46/action.yml#L75) too (that's how I fixed broken dictionary paths), I don't think we need to include `cspell` ones there.

.github/workflows/spelling.yml
```diff
@@ -9,6 +9,7 @@ on:
           The GitHub token to authenticate checkout. Pass in a GitHub personal access token if authenticated submodules
           are used.
 
+    # If updating defaults here also update them in the `spelling` action.
```
BenedekFarkas:
```suggestion
    # When updating defaults here also update them in the `spelling` action.
```

.github/actions/spelling/action.yml
BenedekFarkas: This whole section is influenced by that:
image

The reason why any dictionary shows up in the first place is due to [the default value in check-spelling](https://github.com/check-spelling/check-spelling/blob/ed14e107b2404ceec4db32161bc3cc353b123b46/action.yml#L72), but [the override](https://github.com/Lombiq/GitHub-Actions/blob/e092b644eb21256f4f4057bbc92142201657a615/.github/actions/spelling/action.yml#L166) trims it down to entries that are likely to be relevant to us (and added our own).

But [a comment similar to the one about upgrading to a new version](https://github.com/Lombiq/GitHub-Actions/blob/e092b644eb21256f4f4057bbc92142201657a615/.github/actions/spelling/action.yml#L152-L154) would've probably helped (to check if our override list is still working OK), as it seems that the existing dictionary paths too can break between releases.

Piedone: Got it, thanks.

.github/actions/test-dotnet/Invoke-SolutionTests.ps1
```diff
@@ -38,13 +40,14 @@ $tests = dotnet sln $Solution list |
...
-        # While the test projects are run individually, passing in the solution name via the conventional MSBuild
-        # property allows build customization.
+        # While the test projects are run individually, passing in the solution name and solution dir via the
+        # conventional MSBuild properties allows build customization.
         $switches = @(
             "--configuration:$Configuration"
             '--list-tests'
             "--verbosity:$Verbosity"
             "-p:SolutionName=""$solutionName"""
+            "-p:SolutionDir=""$solutionDirectory"""
```
Piedone, line 50: This was a partial fix for https://lombiq.atlassian.net/browse/CARE-374.

.github/actions/spelling/action.yml
```diff
@@ -125,8 +126,8 @@ runs:
       shell: pwsh
       run: |
         $defaultPrefixes = @{
-          'lombiq-lgha' = 'https://raw.githubusercontent.com/Lombiq/GitHub-Actions/dev/.github/actions/spelling/'
-          'cspell' = 'https://raw.githubusercontent.com/check-spelling/cspell-dicts/v20220816/dictionaries/'
+          'lombiq-lgha' = 'https://raw.githubusercontent.com/Lombiq/GitHub-Actions/${{ github.head_ref }}/.github/actions/spelling/'
```
BenedekFarkas, line 129: This is nice!

BenedekFarkas (approved):

.github/actions/spelling/action.yml
Piedone: Thanks!

BenedekFarkas: I think this change is the issue with the consumer projects' spell checking failing after this was merged to dev. :(
`github.head_ref` is only available in PRs: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
I'll change it to `github.ref_name` for now and test in a separate PR if it has the same effect in a PR as `github.head_ref`.

Piedone: Eh, thanks!

BenedekFarkas: Unfortunately that won't work either, because it's the ref of the triggering workflow, so it won't necessarily match LGHA's default branch name...

Piedone: This is why we can't have nice things. Let's have `dev` here then with a comment.

BenedekFarkas: My thoughts exactly (there is no other suitable workaround), added them on dev already.

PR #264, opened 2023-10-09
OSOE-597: Breaking API change detector during NuGet publishing

### LINKED ISSUE #220: Breaking API change detector during NuGet publishing (OSOE-597)
In the `publish-nuget` action, let's add a feature to detect breaking changes compared to the previous released version, and fail the build if the version number doesn't reflect the breaking change.

- The [package validator](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/baseline-version-validator) feature can be used for this.
- Following [Semantic Versioning](https://semver.org/), only major version increases (e.g. v2.0.0 after v1.2.3) may contain breaking changes. The build should thus fail if the version increase is minor or patch, but a breaking change is detected.
- The baseline version to compare against should be the previous latest released (i.e. not alpha/beta or other prerelease) version of the same lineage. E.g. if v1.2.3 is released it should be compared against v1.2.2 (or any v1.2.x that exists, or if no such exists then v1.x.x), v1.3.0 should be compared against the previous v1.x.x.
- If the major version changes, then no validation should be done since breaking changes are allowed.
- Prerelease versions shouldn't be checked.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-597)

wAsnk:
[OSOE-597](https://lombiq.atlassian.net/browse/OSOE-597)
Fixes #220

[OSOE-597]: https://lombiq.atlassian.net/browse/OSOE-597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -68,11 +68,11 @@ $projects = (Test-Path *.sln) ? (dotnet sln list | Select-Object -Skip 2 | Get-I
 if ($EnablePackageValidation -eq 'True' -And
     $PackageValidationBaselineVersion -And
     !($Version -match '-(alpha|beta|preview|rc)') -And
-    $Version.Substring(0, $version.IndexOf(".")) -le $PackageValidationBaselineVersion.Substring(0, $PackageValidationBaselineVersion.IndexOf(".")))
+    $Version.Substring(0, $version.IndexOf('.')) -le $PackageValidationBaselineVersion.Substring(0, $PackageValidationBaselineVersion.IndexOf('.')))
 {
     Write-Output 'Creating temporary project for baseline NuGet packages.'
     dotnet new classlib -n TempProject
-    cd TempProject
+    Set-Location TempProject
 
     Write-Output 'Installing baseline version NuGet packages.'
     foreach ($project in $projects)
@@ -83,13 +83,14 @@ if ($EnablePackageValidation -eq 'True' -And
     dotnet restore
     Set-Location ..
     Remove-Item -Recurse -Force TempProject
-    $PackageValidationParameters=@(
+    $PackageValidationParameters = @(
         "-p:EnablePackageValidation=$EnablePackageValidation"
         "-p:PackageValidationBaselineVersion=$PackageValidationBaselineVersion"
     )
 }
-else {
-    $PackageValidationParameters=@(
+else
+{
+    $PackageValidationParameters = @(
         "-p:EnablePackageValidation=$EnablePackageValidation"
     )
 }
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -74,7 +74,7 @@ if ($EnablePackageValidation -eq 'True')
     Write-Output "Installing baseline version NuGet packages."
     foreach ($project in $projects)
     {
-        dotnet add TempProject.csproj package $project.BaseName --version $PackageValidationBaselineVersion
+        dotnet add TempProject.csproj package $project.BaseName --version $PackageValidationBaselineVersion --no-restore
     }
 
     dotnet restore
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -102,8 +102,11 @@ foreach ($project in $projects)
     Write-Output "LastMajorVersion: $PackageValidationBaselineVersion"
     if ($EnablePackageValidation -eq 'True')
     {
+        Write-Output "dotnet new classlib -n TempProject"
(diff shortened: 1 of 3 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -178,29 +196,7 @@ runs:
...
-    - name: Create Release
-      # v1.11.2
-      uses: ncipollo/release-action@18eadf9c9b0f226f47f164f5373c6a44f0aae169
-      # This is to prevent creating releases when pushing tags for issue-specific pre-releases like
-      # v4.3.1-alpha.osoe-86.
-      if: "!contains(steps.setup.outputs.publish-version, '-')"
-      with:
-        allowUpdates: true
-        generateReleaseNotes: true
-        artifacts: artifacts/*.nupkg, artifacts/*.snupkg
```
wAsnk: Removed this part until testing, so NuGet packages are not published. When the reviewer is done I'll restore this part.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -14,7 +14,7 @@
     Calls "dotnet pack project.csproj --configuration:Release --warnaserror" on each project.
 #>
 
-param([array] $Arguments)
+param([array] $PackParameters, $EnablePackageValidation, $PackageValidationBaselineVersion, $Version)
```
sarahelsaig: I'd specify the types of all parameters like
```pwsh
param([array] $PackParameters, [bool] $EnablePackageValidation, [string] $PackageValidationBaselineVersion, [string] $Version)
```
Especially for `EnablePackageValidation` so you don' have to interpret it as string on line 68.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -64,6 +64,37 @@ function Get-ProjectProperty
 
 $projects = (Test-Path *.sln) ? (dotnet sln list | Select-Object -Skip 2 | Get-Item) : (Get-ChildItem *.csproj)
 
+# Download baseline version NuGet packages
+if ($EnablePackageValidation -eq 'True' -And
+    $PackageValidationBaselineVersion -And
+    !($Version -match '-(alpha|beta|preview|rc)') -And
```
sarahelsaig: We should border the word from both sides just in case.
```suggestion
    !($Version -match '-(alpha|beta|preview|rc)\.') -And
```

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -64,6 +64,37 @@ function Get-ProjectProperty
...
+    $Version.Substring(0, $version.IndexOf('.')) -le $PackageValidationBaselineVersion.Substring(0, $PackageValidationBaselineVersion.IndexOf('.')))
```
sarahelsaig: The second `$version` is not capitalized.

sarahelsaig: Both `.Substring()` calls can be simplified using `$Version.Split('.')[0]` instead.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -64,6 +64,37 @@ function Get-ProjectProperty
...
+{
+    Write-Output 'Creating temporary project for baseline NuGet packages.'
+    dotnet new classlib -n TempProject
+    Set-Location TempProject
```
sarahelsaig: When you enter a location and want to leave it afterwards use `Push-Location path; ...; Pop-Location` instead of `Set-Location path; ...; Set-Location ..` - it communicates the intent better.

sarahelsaig (changes requested):

sarahelsaig:
Approved. Please revert https://github.com/Lombiq/GitHub-Actions/pull/264#discussion_r1350186544 and then you can go ahead and merge too.

wAsnk:
> Approved. Please revert [#264 (comment)](https://github.com/Lombiq/GitHub-Actions/pull/264#discussion_r1350186544) and then you can go ahead and merge too.

Please approve the PR otherwise the Validate GitHub Actions Refs check will fail.

wAsnk:
@DAud-IcI

sarahelsaig (approved):
> Please approve the PR otherwise the Validate GitHub Actions Refs check will fail.

Neat!

sarahelsaig:
`validate-gha-refs` still fails though.
```
Error: .github/workflows/publish-nuget.yml#L78 - called GitHub Action does not match expected Ref 'issue/OSOE-597'.
Error: Check Job Summary for more details on which GitHub Actions Refs do not match 'issue/OSOE-597'.
Error: Process completed with exit code 1.
```

wAsnk:
This was fun!

PR #268, opened 2023-10-16
OSOE-707: NuGet package validation fails if new module is added in new version

### LINKED ISSUE #267: NuGet package validation fails if new module is added in new version (OSOE-707)
When a new module is added to the repository that already has published NuGet packages, the NuGet package validation will fail because it won't find the new module with the baseline NuGet version as it is a totally new module.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-707)

wAsnk:
[OSOE-707](https://lombiq.atlassian.net/browse/OSOE-707)
Fixes #267

[OSOE-707]: https://lombiq.atlassian.net/browse/OSOE-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/New-NuGetPackage.ps1
+++ b/.github/actions/publish-nuget/New-NuGetPackage.ps1
@@ -117,7 +117,7 @@ foreach ($project in $projects)
 
         if ($LASTEXITCODE -ne 0)
         {
-            Write-Output '::warning::Package version couldn't be added, thus package validation to baseline version won't be done.'
+            Write-Output "::warning:: Package version couldn't be added, thus package validation to baseline version won't be done."
             dotnet remove TempProject.csproj package $project.BaseName --version $PackageValidationBaselineVersion
             $PackageValidationParameters = @(
                 "-p:EnablePackageValidation=$EnablePackageValidation"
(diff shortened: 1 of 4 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -200,26 +200,3 @@ runs:
           )}
 
         New-NuGetPackage @parameters
-
```
wAsnk: This was removed only for testing so it won't publish to NuGet, if reviewed and approved, I will restore it.

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -126,6 +100,41 @@ foreach ($project in $projects)
...
+        Write-Output 'Creating temporary project for baseline NuGet package.'
+        dotnet new classlib -n TempProject
+        Push-Location TempProject
+
+        Write-Output 'Installing baseline version NuGet package.'
+        dotnet add TempProject.csproj package $project.BaseName --version $PackageValidationBaselineVersion
+
+        if ($LASTEXITCODE -ne 0)
+        {
+            Write-Output "Package version couldn't be added, thus package validation to baseline version won't be done."
```
sarahelsaig: This seems fairly exceptional so I think it should appear on the workflow summary page too:
```suggestion
            Write-Output '::warning::Package version couldn't be added, thus package validation to baseline version won't be done.'
```

Also constant string literals should be bordered with single quotes, see [AvoidUsingDoubleQuotesForConstantString](https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingdoublequotesforconstantstring?view=ps-modules).

sarahelsaig (changes requested):

sarahelsaig (approved):

PR #276, opened 2023-11-10
OSOE-723: Upgrading dotnet consolidate to v4.1.0 and surfacing new excludedVersionsRegex config

Piedone:
OSOE-723

### DIFF
--- a/.github/actions/verify-dotnet-consolidation/action.yml
+++ b/.github/actions/verify-dotnet-consolidation/action.yml
@@ -27,7 +27,10 @@ runs:
       with:
         name: dotnet-consolidate
         version: 4.1.0
+
     - name: Verify consolidation status
       shell: pwsh
       working-directory: ${{ inputs.directory }}
-      run: dotnet consolidate -s (Get-ChildItem ${{ inputs.solution-path }}).FullName --excludedVersionsRegex ${{ inputs.exclude-version-regex }}
+      env:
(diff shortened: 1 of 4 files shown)


.github/actions/verify-dotnet-consolidation/action.yml
```diff
@@ -22,8 +26,8 @@ runs:
       uses: Lombiq/GitHub-Actions/.github/actions/install-dotnet-tool@dev
       with:
         name: dotnet-consolidate
-        version: 2.0.0
+        version: 4.1.0
     - name: Verify consolidation status
       shell: pwsh
       working-directory: ${{ inputs.directory }}
-      run: dotnet consolidate -s (Get-ChildItem ${{ inputs.solution-path }}).FullName
+      run: dotnet consolidate -s (Get-ChildItem ${{ inputs.solution-path }}).FullName --excludedVersionsRegex ${{ inputs.exclude-version-regex }}
```
sarahelsaig: Are you sure this will work with more complex expressions? The `${{ }}` just substitutes, so I think it will break with anything containing shell operators such as `*`. I'd rather go for
```
env:
  ExcludeVersionRegex: ${{ inputs.exclude-version-regex }}
```

sarahelsaig (changes requested):

.github/actions/verify-dotnet-consolidation/action.yml
Piedone: Yeah, good point.

PR #287, opened 2023-11-24
OSOE-729: Prevent Azure-related workflows running at the same time when interacting with the same environment(s)

### LINKED ISSUE #281: Prevent Azure-related workflows running at the same time when interacting with the same environment(s) (OSOE-729)
For example, running `deploy-to-azure-app-service` and `reset-azure-environment` against the same environment will cause them to step on each other's toes (and fail or do something unexpected), same goes for any of these two and `swap-azure-web-app-slots` (also fix the O1 variants).

Update the [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) group names to be dynamic by using the slot name parameters.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-729)

BenedekFarkas:
[OSOE-729](https://lombiq.atlassian.net/browse/OSOE-729)
Fixes #281

[OSOE-729]: https://lombiq.atlassian.net/browse/OSOE-729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Docs/Workflows/Index.md
BenedekFarkas, line 1: The only content change in this file is the categories section.

Docs/Workflows/BuildDotNetCoreOrchardCore.md
BenedekFarkas, line 1: Just factored out from the earlier Workflows.md and changed the second-level titles.

Docs/Workflows/AzureHosting.md
BenedekFarkas, line 1: Just factored out from the earlier Workflows.md and changed the second-level titles. Also moved common notes to the primary section.

Docs/Actions.md
BenedekFarkas, line 1: Text didn't change, just added categories.

Docs/Workflows/Productivity.md
BenedekFarkas, line 1: Just factored out from the earlier Workflows.md and changed the second-level titles.

Docs/Workflows/Index.md
Piedone: This breaks incoming links (we also have those in our wiki). This is necessary, but at that point, let's make it future proof: put all the workflows' docs into their own pages, and then you can elevate the content of this page (with the general notes and links to workflows' pages) to the Readme, so people have to dig less.

Psichorex: Looks good to me after the refactoring but I @Piedone please approve if you are fine as its your request.

Piedone: I'm not fine, hence my comment :).

Psichorex: Wait didn't this got addressed by the factored out .mds?
Edit: I see now, my brain got stuck at the new .mds.

BenedekFarkas: Re-added the original Workflow.md with the complete Workflows "table of contents", I think that's a good middle-ground between information accessibility and density, I'd rather not put all this into the Readme. Folder structure is according to the categorization in Workflow.md.

Piedone: Look good. The link in the Readme is broken BTW.

BenedekFarkas: Yeah, I noticed, went around just now and checked/fixed every link.

Psichorex: All the links seems to be fine, at least on my side so probably overall aswell. Went throught manually on most of them and I yet to find any broken ones.

Psichorex (approved):

Psichorex:
I will merge if @Piedone approves. Everything seems fine. Approved by me.

Piedone:
It's only up to you.

Psichorex:
> It's only up to you.

I asked because you had an unresolved comment containing a request for which I didn't see a clear indication that it was resolved.

Docs/Workflows/Index.md
Psichorex: > Yeah, I noticed, went around just now and checked/fixed every link.

From this reply I thought that all links are fine now but for me SpellChecking.md has broken links.
https://github.com/Lombiq/GitHub-Actions/blob/763adec7ca1b54e16985554d97b88b253196b96f/Docs/Workflows/Productivity/SpellChecking.md#:~:text=proxied%20by%20the-,spelling%20action,-in%20this%20repository

Piedone:
I understand, but it's still up to you, me doing a second review of this PR is unnecessary.

Docs/Workflows/Index.md
BenedekFarkas: Fixed!

Psichorex (approved):

PR #290, opened 2023-11-28
OSOE-743: Cancel ongoing runs of long-running workflows under a PR when new changes are pushed

### LINKED ISSUE #288: Cancel ongoing runs of long-running workflows under a PR when new changes are pushed (OSOE-743)
This is particularly interesting for long-running workflows (project/solution builds) to reduce resource consumption, but small things add up for the shorter ones (e.g., spell-checking) as well. Example: When a commit in a PR triggers the solution build and another commit is pushed while the previous one's build is still running, we can optionally cancel the now outdated build using [concurrency groups](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency):

```yaml
concurrency:
  group: ${{ inputs.cancel-in-progress-for-this-pr == 'true' && github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.ref) || github.run_id }}
  cancel-in-progress: ${{ inputs.cancel-in-progress-for-this-pr }}

on:
  workflow_call:
    inputs:
      cancel-in-progress-for-this-pr:
        type: string
        default: 'false'
```

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-743)

BenedekFarkas:
Fixes https://github.com/Lombiq/GitHub-Actions/issues/288.


.github/workflows/validate-this-gha-refs.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - dev
-  pull_request:
+  workflow_dispatch:
```
Piedone: Why is this and in `validate-this-pull-request` necessary?

Docs/Workflows.md
```diff
@@ -8,6 +8,7 @@ These workflows can be invoked from a step from any other repository's workflow.
...
 - To add the workflows to a project create a folder in the root of the repository that will call them, e.g. _.github/workflows/build-and-test.yml_ and/or _.github/workflows/publish-nuget.yml_. The examples below are for such files.
 - If you use these workflows with a self-hosted runner, then you'll need to fork this repository under the organization while keeping [these rules](https://docs.github.com/en/actions/using-workflows/reusing-workflows#access-to-reusable-workflows) in mind. Then, you need to enable workflows for the fork under its Actions tab (you'll see a big button for this). If you don't do the latter step, you'll get a "workflow was not found" error. Then you can also disable the `spelling-this-repo` and `validate-this-gha-refs` workflows, not to run them unnecessarily if you sync the fork with the original repo.<!--#spell-check-ignore-line-->
+- Some of the long-running workflows (for example solution builds and spell-checking) also support the `cancel-in-progress-for-this-pr` parameter. When it's set to `'true'`, will cancel the already running workflow run or job (e.g., if the job uses strategy matrix to run on different machine types). Default is `'false'`.
```
```suggestion
- Some of the long-running workflows (for example solution builds and spell-checking) also support the `cancel-in-progress-for-this-pr` parameter. When it's set to `'true'`, it will cancel the already running workflow run or job (e.g., if the job uses strategy matrix to run on different machine types) when a newer workflow is started for a given pull request. You can use this to conserve computing resources (like paid GitHub Actions minutes) by not letting frequent commits under a PR each run their own build, for example, since only the most recent one will be run to completion. The default is `'false'`.
```

Piedone: I'd keep the default on `true`. Similar to what we do with `cancel-workflow-on-failure` by providing the default that saves you time and money, we should do it here. Then, you can selectively disable it if you need it for a given task.

.github/workflows/build-and-test-dotnet.yml
```diff
@@ -10,8 +10,14 @@ on:
           are used.
 
     inputs:
+      cancel-in-progress-for-this-pr:
+        description: >
+          When set to "true", it will cancel the already running workflow for this PR. See the concurrency settings of
+          the job for more details.
```
Piedone: Also for all the others. "PR" is not an obvious abbreviation if you're not already thinking in terms of pull requests, and these workflows aren't just executed under PRs.

```suggestion
          When set to "true", it will cancel the already running workflow for this pull request. See the concurrency
          settings of the job for more details.
```

.github/workflows/build-and-test-dotnet.yml
```diff
@@ -115,7 +121,7 @@ on:
       dotnet-consolidation-exclude-version-regex:
         type: string
         required: false
-        default: "-alpha"
+        default: '-alpha'
```
Piedone, line 124: Why this change? When using single quotes, the samples in the descriptions (like <<When set to "true">>) should also use single quotes.

Also in the other workflows.

Related to https://github.com/Lombiq/NodeJs-Extensions/issues/11.

Docs/Workflows.md
Piedone: When will the workflow be canceled and when just the job, exactly? E.g., what will happen with a workflow like OSOCE's, that calls the `build-and-test-orchard-core` and `spelling` reusable workflows, where the former uses a matrix? Should one only use these configs, or is there any scenario (when only calling such reusable workflows from your own workflow) when you'd put `cancel-in-progress: true` in your own workflow instead? Add this to the docs too.

Piedone (changes requested):

.github/workflows/validate-this-gha-refs.yml
BenedekFarkas: Just temporary!

Docs/Workflows.md
BenedekFarkas: Agree, OK!

.github/workflows/build-and-test-dotnet.yml
BenedekFarkas: It was autoformatting (`"yaml.format.singleQuote": true,` setting helps with this for VS Code) to match all the other parameters. Example values in descriptions are surrounded by doublequotes (at least in the workflows I modified, so no change needed there, because it's not machine-processed). There are a few cases of backticks too, but that just makes it harder to type/read, because it's not rendered.

Piedone: I'm OK with the single quote style (since in `if`s it's required anyway; though changing everything instead of just certain parameters would be better) but if we say that strings should be surrounded with single quotes then examples of strings should use that too.

BenedekFarkas: I intended to go through each file and add fix formatting for every file, but that would be a lot of stuff in the diff that is not relevant to this PR, so I kept it within the boundaries of the files I'm working on to make it consistent at least there.

Since we're working on automating YAML linting, we'll have to change a lot of stuff soon anyway.

However, default values (for workflows, jobs, parameters) don't need any surrounding apostrophes or double quotes at all, so we should get rid of those completely (and enforce this).

Text intended for humans only (i.e., not processed/rendered) should use double quotes.

Docs/Workflows.md
BenedekFarkas: What I wrote was unclear, because cancellation is applied to the jobs only, so I simplified what's described here. There are multiple different cases and situations to consider, but I don't think this is the place to explain them all. However, the workflows that include this parameter supply their own explanation as in-line comments at the concurrency configuration (including how and why it's configured at the workflow or job level).

.github/workflows/build-and-test-dotnet.yml
Piedone: Yeah, I wouldn't want all such styling changes in this PR either.

PR #314, opened 2024-01-26
OSOE-769: Trigger validate-pull-request on pull_request_target

### LINKED ISSUE #304: Trigger validate-pull-request on pull_request_target (OSOE-769)
The [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) trigger is better than `pull_request`, since it also works for forks, and without approval. That's what we actually need.

Then [this line](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/check-merge-conflict/action.yml#L11) in `check-merge-conflict` can be adjusted too.

Same for `validate-submodule-pull-request`.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-769)

Piedone:
Fixes #304
OSOE-769

### DIFF
--- a/.github/actions/check-merge-conflict/action.yml
+++ b/.github/actions/check-merge-conflict/action.yml
@@ -5,10 +5,6 @@ runs:
   using: "composite"
   steps:
     - name: Check for merge conflict
-      # If this is run for a PR opened from a fork (not from a branch in the given repo) then "Resource not accessible
-      # by integration" errors can happen. Thus, not running it for forks during the pull_request event (but it can
-      # run during push, for example). See https://github.com/Lombiq/GitHub-Actions/issues/72 for details.
-      if: github.event.pull_request == '' || github.event.pull_request.head.repo.fork == false
       # releases/2.1.0
       uses: eps1lon/actions-label-merge-conflict@92924ba33a60e436034b0ac3838de523bf7df071
       with:
--- a/.github/actions/update-github-issue-and-pull-request/action.yml
+++ b/.github/actions/update-github-issue-and-pull-request/action.yml
@@ -1,5 +1,7 @@
 name: Update Github Issue and Pull Request
-description: Prefixes the PR title with a Jira issue key, adds the Jira key and the issue key to the body if not present, and also assign the issue to PR author.
+description: |
+  Prefixes the PR title with a Jira issue key, adds the Jira key and the issue key to the body if not present, and also
+  assigns the issue to the PR author.
 
 runs:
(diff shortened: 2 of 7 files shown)


BenedekFarkas:
Wouldn't these changes prevent our team members from testing their own PRs' changes to the `validate(-submodule)-pull-request` workflow?

Piedone:
Yes, for testing those you'd need to change the trigger. There's unfortunately no way around this.

BenedekFarkas:
OK, I'm trying to picture in my head what the effect of this will be exactly. If all the triggers are updated, including `validate-pull-request` in OSOCE and `validate-this-pull-request` in LGHA, then changes made to those can't be tested in a PR at all. **Is that right?**

For example, if I work on the `validate-pull-request` reusable workflow, I'd also update the branch references in the workflows that call it, including in OSOCE in LGHA, but this way OSOCE and LGHA will still run on the target (usually default branch).

Can this be circumvented by making changes (with malicious intent to extract a secret/token) by opening a second PR that targets my first one with the bad code?

Piedone:
Those changes won't be possible to test without doing anything, but will be possible to test after changing the workflow's trigger from `pull_request_target` to `pull_request` (and possibly [this](https://github.com/Lombiq/GitHub-Actions/blob/b34f78ec10f1870bb0777819daca21954ef6e36a/.github/workflows/validate-pull-request.yml#L24) and [this](https://github.com/Lombiq/GitHub-Actions/blob/b34f78ec10f1870bb0777819daca21954ef6e36a/.github/workflows/validate-submodule-pull-request.yml#L28) too, but you'll need to change `opened` here anyway unless you want to open a new PR for every test run).

I guess so, yes, an external actor can open a PR targeting an issue branch that has the `pull_request` trigger like this. However, just as with PRs today, the workflow there wouldn't run until one of our team members goes there and approves the run.

The point of these changes is to allow running Validate Pull Request right when opening the PR (i.e. without anybody clicking the "Approve and run" button there) even for external contributors. This wouldn't be safely possible with `pull_request` (since that would run the workflow code in the PR) and thus GitHub doesn't allow it. It is safe though with `pull_request_target` (since that runs the workflow code of the target branch, but only if the target branch already uses `pull_request_target`).

BenedekFarkas:
> but only if the target branch already uses `pull_request_target`

OK, thanks, this was the missing piece for me.

BenedekFarkas (approved):

PR #325, opened 2024-02-09
OSOE-735: Introduce versioning and releases

### LINKED ISSUE #284: Introduce versioning and releases (OSOE-735)
See https://github.com/orgs/community/discussions/70237 for a detailed problem statement and the solution we need to introduce here.

In a nutshell, due to how references between workflows and actions work in GHA, LGHA reusable workflows can't be used in a versioned way, you have to use the rolling releases of `@dev` (using actions directly is not affected).

To overcome this, we need to have explicit releases, like `v1.0` and `v1.1`. We'd add tags for these in Git, and when we do that, all the branch specifiers in the repo, which are currently `@dev`, would be replaced with `v1.0` automatically.

We could have special branches, like `release/v1.0`, where we'd merge from `dev`. A workflow would pick up commits to any such branches, replace branch specifiers in a commit in the branch, add a tag, and publish a release on GitHub. Special care is needed for the `spelling` action, due to the additional branch reference [here](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/spelling/action.yml#L133).

Some alternative release process may be suitable too, so don't keep it to yourself, but ultimately, we need to a) be able to tag releases and b) everything after that should happen automatically.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-735)

davidpuplava:
[OSOE-735](https://lombiq.atlassian.net/browse/OSOE-735)
Fixes #284

[OSOE-735]: https://lombiq.atlassian.net/browse/OSOE-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

EDIT:
Adding additional information to this summary based on the feedback from coderabbitai. 

Regarding the `spelling` action, I've added a `additional-pattern-include-list` option to account for special cases such as referencing files directly from the dev branch.

Regarding alternate release process, the approach that we're taking here seems to be consistent with how other organizations version GitHub Actions. One difference is that well know actions such as `actions/checkout` are placed into their own repository. Considering how many different actions and workflows exist in LGHA, splitting them into different repositories is impractical, and not a very good developer experience.

### Explanation of this Change

The process for version LGHA is as follows:
1. Create a new release branch using the following convention `release/v1.0` where the v1.0 is which every version you want to release.
2. Push your `release/v1.0` branch
3. A new workflow [tag-and-release](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-and-release.yml) will trigger based on the push of the release branch and perform the following steps.
4. Automatically take the branch name `release/v1.0` and remove the `release/` to get the version tag `v1.0`
5. Execute a new action `set-gha-refs` that will scan GitHub Actions and Workflows looking for `@dev` refs to update
6. This new action will also look for generic patterns in the [additional-pattern-include-list](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/actions/set-gha-refs/action.yml#L24) which allows for special considerations such as for the [spelling action explicit file reference](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/actions/spelling/action.yml#L133).
7. The workflow and action files will be updated to use the new `v1.0` reference.
8. These file changes will automatically be committed to the `release/v1.0`
9. A new `v1.0` tag will be created.
10. The branch and tag will be pushed to the repository
11. The v1.0 tag will be `--force` pushed to the repo to update and existing tags that will now be outdated.

## Security Considerations
The new `tag-and-release` workflow depends on a personal access token that should be configured in a [secret named `TAG_AND_RELEASE`](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-and-release.yml#L18). This token is used in place of the default GITHUB_TOKEN that is typically used in workflows. The reason for this is because the token must have the `workflow` permission granted in order to push changes to GitHub Action files. By default, GITHUB_TOKEN does not have this permission.


.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,35 @@
...
+    [String[]] $PathIncludeList,
+    [String[]] $FileIncludeList,
+    [String] $ExpectedRef
+)
+
+if ($CalledRepoBaseIncludeList.Count -eq 0)
+{
+    Write-Output '::warning file=Check-Called-GHA-refs.ps1,line10::CalledRepoBaseIncludeList is empty which is unexpected. If this was intentional, you can ignore this warning.'
+    exit 0 # Nothing to check because array is empty.
+}
```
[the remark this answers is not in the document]
davidpuplava: This is now fixed by moving the rest of the code into the else block.

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,35 @@
...
+
+$CalledRepoBaseIncludeList = $CalledRepoBaseIncludeList.ForEach({ 'uses:\s*' + $PSItem + '(.*)@(.*)' })
```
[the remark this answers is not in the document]
davidpuplava: Based on [GitHub's naming restrictions](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags), I don't see any overlap with regex special characters, so I don't think this will be an issue.

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,35 @@
...
+
+$matchedRefs = Get-ChildItem -Path $PathIncludeList -Include $FileIncludeList -Force -Recurse |
+    Select-String -Pattern $CalledRepoBaseIncludeList
```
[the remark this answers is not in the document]
davidpuplava: In this context, any file access errors should bubble up to the stop and allow the workflow run to handle the error.

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
```diff
@@ -0,0 +1,35 @@
...
+    foreach ($matched in $matchedRefs)
+    {
+        $oldline = $matched.Line
+        $newline = $matched.Line -Replace '@(.*)', "@$ExpectedRef"
+        Write-Output $oldline
+        Write-Output $newline
+
+        $filename = $matched.RelativePath($pwd)
+
+        (Get-Content $filename).Replace($oldline, $newline) | Set-Content $filename
```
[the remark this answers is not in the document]
davidpuplava: Since the modified files are under version control, no data loss is likely nor are backups needed.

.github/workflows/tag-and-release.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Update LGHA Refs
+    runs-on: ubuntu-latest
+    permissions:
+      # Give the default GITHUB_TOKEN write permission to commit and push the changed files back to the repository.
+      contents: write
+    steps:
+      - name: Checkout Repository
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.TAG_AND_RELEASE }}
```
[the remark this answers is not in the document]
davidpuplava: Agreed, but the current process is to target the `@dev` branch. Interestingly enough, this PR is a stop toward changing that process.

.github/workflows/tag-and-release.yml
```diff
@@ -0,0 +1,39 @@
...
+          $refname = "${{ github.ref_name }}"
+          $tagname = $refname -replace 'release/', ''
+          $output = "tagname=$tagname"
+          Write-Output "output=$output"
+          $output >> $env:GITHUB_OUTPUT
+      
+      - name: Set Ref for GitHub Actions and Workflows
+        uses: Lombiq/GitHub-Actions/.github/actions/set-gha-refs@issue/OSOE-735
+        with:
+          expected-ref: '${{ steps.determine-tag.outputs.tagname }}'
```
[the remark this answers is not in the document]
davidpuplava: Same as a above, this is expected.

.github/actions/set-gha-refs/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Set References
+      shell: pwsh
+      run: |
+        $params = @{
+            PathIncludeList = ${{ inputs.path-include-list }}
+            FileIncludeList = ${{ inputs.file-include-list }}
+            CalledRepoBaseIncludeList = ${{ inputs.called-repo-base-include-list }}
+            ExpectedRef = "${{ inputs.expected-ref }}"
+        }
+        Set-Called-GHA-Refs @params
```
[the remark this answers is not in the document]
davidpuplava: PowerShell has [protection against some common injection techniques](https://devblogs.microsoft.com/powershell/protecting-against-malicious-code-injection/) and the underlying script does not use Invoke-Expression so additional sanitize is not necessary here.

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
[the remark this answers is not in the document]
davidpuplava: Here is the else block: https://github.com/Lombiq/GitHub-Actions/pull/325/files#diff-9efb0b66768e6e51063c3d4c02f8b392c38055126adfb32033d4492696415cc5R12-R40

.github/workflows/tag-version-this-repo.yml
```diff
@@ -0,0 +1,13 @@
+name: Tag Version (this repo)
```
Piedone, line 1: Add docs about this under https://github.com/Lombiq/GitHub-Actions?tab=readme-ov-file#contributing-and-support as well, please. Basically, what you've written in the PR description.

BenedekFarkas: Please also clarify which steps are manual and which are automatic. For example steps 1-2 are manual for sure, 3-4 are automatic, but the wording is unclear for me for step 5.

Also, it looks like to me that contrary to
> The new tag-and-release workflow depends on a personal access token that should be configured in a [secret named TAG_AND_RELEASE](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-and-release.yml#L18).

the name of the token doesn't matter, since you're passing it in as a parameter.

.github/actions/set-gha-refs/action.yml
BenedekFarkas, line 1: Please use Rewrap to wrap long description lines at 120 characters for each YAML file.

.github/actions/set-gha-refs/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    description: >
+      PowerShell string array of file name patterns to include when evaluating GHA files, e.g. '@("*.yml")' or '@("*.yml", "*.yaml")'. The
+      parameter must be a PowerShell string array.
+  called-repo-base-include-list:
+    required: false
+    default: '@("${{ github.repository }}")'
+    description: >
+      PowerShell string array of repository base URLs to include when evaluating called GHA Workflows and Actions, e.g '@("Lombiq/GitHub-Actions")' or
+      '@("Lombiq/GitHub-Actions", "Lombiq/Open-Source-Orchard-Core-Extensions")'. The parameter must be a PowerShell string array.
+  additional-pattern-include-list:
```
BenedekFarkas, line 25: This is nice!

.github/workflows/tag-version.yml
BenedekFarkas, line 1: The wrapper workflow should surface the same parameters as the action it wraps, see spelling for example.

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          $refname = "${{ github.ref_name }}"
+          $tagname = $refname -replace 'release/', ''
+          $output = "tagname=$tagname"
+          Write-Output "output=$output"
+          $output >> $env:GITHUB_OUTPUT
+      - name: Set Ref for GitHub Actions and Workflows
+        uses: Lombiq/GitHub-Actions/.github/actions/set-gha-refs@issue/OSOE-735
+        with:
+          additional-pattern-include-list: '@("https://raw.githubusercontent.com/Lombiq/GitHub-Actions/(?<ref>[aA-zZ,0-9,-,\.,/,_]*)/.github")'
```
BenedekFarkas: When the parameter list is updated to include everything from the action, this parameter value should be passed in from the `tag-version-this-repo` workflow.

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,51 @@
...
+          expected-ref: '${{ steps.determine-tag.outputs.tagname }}'
+      - name: Commit and Push Changes
+        uses: stefanzweifel/git-auto-commit-action@v5
```
BenedekFarkas: External actions/workflows should always be pinned to a commit hash for security reasons with the associated version number displayed in a comment, for example:
`uses: check-spelling/check-spelling@00c989c97749eb0cb2d256bdc55ac61b0096c6d3 # v0.0.22`

.github/workflows/tag-version-this-repo.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  push:
+    branches:
+      - release/**
+
+jobs:
+  run:
+    name: Tag Version Automation
+    if: github.event.pusher.name != 'LombiqBot'
```
BenedekFarkas, line 11: What's the larger context for having to include this specific condition?

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  run:
+    name: Set GitHub Action/Workflow to Version Tag
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.TAG_VERSION_TOKEN }}
+      - name: Determine Version Tag Name from Branch Name
```
BenedekFarkas, line 78: Add an empty line between each step for readability.

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: determine-tag
+        shell: pwsh
+        run: |
+          $refname = "${{ github.ref_name }}"
+          $tagname = $refname -replace 'release/', ''
+          $output = "tagname=$tagname"
+          Write-Output "output=$output"
+          $output >> $env:GITHUB_OUTPUT
```
BenedekFarkas: Use the Set-GitHubOutput script from the root Scripts folder of this repo (needs a "Setup Scripts"-like step, see the spelling action for example).

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          additional-pattern-include-list: '@("https://raw.githubusercontent.com/Lombiq/GitHub-Actions/(?<ref>[aA-zZ,0-9,-,\.,/,_]*)/.github")'
+          expected-ref: '${{ steps.determine-tag.outputs.tagname }}'
+      - name: Commit and Push Changes
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'Set GitHub Actions/Workflows to tag ${{ steps.determine-tag.outputs.tagname }}'
+          tagging_message: ${{ steps.determine-tag.outputs.tagname }}
+      - name: Force Push Tag
+        continue-on-error: true
```
BenedekFarkas, line 100: In case of an error, the user (developer) will still be notified about it?

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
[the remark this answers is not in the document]
BenedekFarkas: Here is the whole script:

```
param(
    [String[]] $CalledRepoBaseIncludeList,
    [String[]] $AdditionalPatternIncludeList,
    [String[]] $PathIncludeList,
    [String[]] $FileIncludeList,
    [String] $ExpectedRef
)

if ($CalledRepoBaseIncludeList.Count -eq 0)
{
    Write-Output '::warning file=Check-Called-GHA-refs.ps1,line10::CalledRepoBaseIncludeList is empty which is unexpected. If this was intentional, you can ignore this warning.'
}
else
{
    $CalledRepoBaseIncludeList = $CalledRepoBaseIncludeList.ForEach({ 'uses:\s*' + $PSItem + '.*@(?<ref>[aA-zZ,0-9,-,\.,/,_]*)' })

    $matchedRefs = Get-ChildItem -Path $PathIncludeList -Include $FileIncludeList -Force -Recurse |
        Select-String -Pattern $CalledRepoBaseIncludeList

    $additionalRefs = Get-ChildItem -Path $PathIncludeList -Include $FileIncludeList -Force -Recurse |
        Select-String -Pattern $AdditionalPatternIncludeList

    $matchedRefs = $matchedRefs + $additionalRefs

    if ($matchedRefs.Count -gt 0)
    {
        foreach ($matched in $matchedRefs)
        {
            $oldline = $matched.Line
            $newline = $matched.Line -Replace $matched.Matches[0].Groups['ref'].Value, "$ExpectedRef"

            if ($oldine -ne $newline)
            {
                Write-Output "$oldline => $newline"

                $filename = $matched.RelativePath($pwd)
                $linenumber = $mismatch.LineNumber
                $title = "GHA Ref pinned to '$ExpectedRef'"

                (Get-Content $filename).Replace($oldline, $newline) | Set-Content $filename

                Write-Output "::notice file=$filename,line=$linenumber,title=$title::GHA Ref changed to '$ExpectedRef'"
            }
        }
    }
}
```

BenedekFarkas (changes requested):
I tested this in a fork and tagging worked great!

Question: Shouldn't workflow/action references throughout the whole repo be updated? [See this commit](https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/commit/28ce10657a408386353082c78bf24966e1ac3078) that was triggered by the release branch and created the tag.

For example, if I use a workflow from the v1.0 tag, then it will call other workflows and actions still from the dev branch.

.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1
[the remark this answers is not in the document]
BenedekFarkas: @coderabbitai you asked to see the contents `.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1` script that includes the else block - see my previous comment so you can continue your review.

.github/workflows/tag-version-this-repo.yml
davidpuplava: Thanks @Piedone, I'll modify the docs. I'll review those steps @BenedekFarkas - I think I may have changed to using a passed in secret after I wrote the PR steps and forgot to update.

BenedekFarkas:
@davidpuplava just a reminder, see my review comments.

davidpuplava:
Thanks @BenedekFarkas - I will look at this today

.github/workflows/tag-version-this-repo.yml
davidpuplava: Added new section to readme for [Versioning, Tags and Releases](https://github.com/Lombiq/GitHub-Actions/tree/issue/OSOE-735?tab=readme-ov-file#versioning-tags-and-releases).

.github/actions/set-gha-refs/action.yml
davidpuplava: Thank you Rewrap is a very nice VS extension! Done, see 4ee3af6f7ff65b01825f58cdb72a28f360037465.

.github/actions/set-gha-refs/action.yml
davidpuplava: Thank you - I took this idea from the [Verify GitHub Actions References](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/verify-gha-refs/action.yml) action. If I remember correctly, the reason for a list was to allow for the calling workflow to be able to optionally pass in additional `.github` directories. For example, if you wanted to include a specific submodule (for whatever reason) when verifying any of the GHA `@dev` refs.

The `additional-pattern-include-list` was necessary to handle the [special circumstance](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/spelling/action.yml#L133) with the spelling action that @Piedone mentioned in issue #284.

.github/workflows/tag-version.yml
davidpuplava: Agreed. Refactored in 66de8664c3c3c71430ef612a3758f5021e3bfd70.

.github/workflows/tag-version.yml
davidpuplava: Agreed. Included in the refactor in 66de8664c3c3c71430ef612a3758f5021e3bfd70.

.github/workflows/tag-version.yml
davidpuplava: Fixed in 08bf3cfe296447e0bd20724c722e057003168129.

.github/workflows/tag-version-this-repo.yml
davidpuplava: Good question. After the `@release/v1.0` refs are updated to `@v1.0` in the GHA files, they are automatically committed on the `release/v1.0` branch and pushed (by the `LombiqBot` user). And because the workflow is triggered on the `push` event to any `release/**` branch, it kicks off a 2nd workflow run.

This specific condition exits early from that 2nd workflow run.

One important question to ask here is, do we really want this commit on the `release/v1.0` branch and then have it pushed?

One problem I see is any additional commits pushed to the `release/v1.0` branch, fixing a small bug for example, may be challenging. I think you'll get an out of date error when pushing and you'll need to pull the latest commit, All the refs will point to the new `@v1.0` tag instead of the `@release/v1.0` branch. So you'll have to change them all back before pushing again.

Maybe the solution is to put the commit on a different branch like `main` or a new one like `release-asset/v1.0`. Then at the very least, the 2nd run of the workflow won't be triggered on the push to the `release/v1.0` branch.

.github/workflows/tag-version.yml
davidpuplava: Yep, done in eed4713f3d49a78fa55a374339335a28312d6fd5.

.github/workflows/tag-version.yml
davidpuplava: Refactored in ab3535ec8ed3ab95a470cd74b89666d7331a0903.

.github/workflows/tag-version.yml
davidpuplava: > In case of an error, the user (developer) will still be notified about it?

No, not in this case. This is here for a very narrow edge case where the tag is created but the automatic commit from the prior [Commit and Push Changes](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version.yml#L92) step fails to push the commit up to the remote repository. If you push a tag to a remote but the underlying tag is not in the remote, then `git push` returns an error code and stops the workflow.

davidpuplava:
> @BenedekFarkas Question: Shouldn't workflow/action references throughout the whole repo be updated? [See this commit](https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/commit/28ce10657a408386353082c78bf24966e1ac3078) that was triggered by the release branch and created the tag.

Yes, the reference should be updated in the entire repository. Let me try and replicate this by forking the repo and trying to create a release in my own remote.

I'll have to check but I think this might be because it's running in a fork. But it may no longer be an issue now that the action inputs were surfaced up to the calling workflows.

.github/workflows/tag-version.yml
BenedekFarkas: Gotcha, thanks!

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,102 @@
...
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.TAG_VERSION_TOKEN }}
+
+      - name: Determine Version Tag Name from Branch Name
+        id: determine-tag
+        run: |
+          $refname = "${{ inputs.expected-ref }}"
+          $tagname = $refname -replace 'release/', ''
+          Set-GitHubOutput -Key 'output' -Value $tagname
```
BenedekFarkas:
```suggestion
          Set-GitHubOutput -Key 'tagname' -Value $tagname
```

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Checkout Repository
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.TAG_VERSION_TOKEN }}
+
+      - name: Determine Version Tag Name from Branch Name
+        id: determine-tag
+        run: |
+          $refname = "${{ inputs.expected-ref }}"
+          $tagname = $refname -replace 'release/', ''
```
BenedekFarkas:
```suggestion
          $tagname = "${{ inputs.expected-ref }}" -replace 'release/', ''
```

BenedekFarkas (changes requested):

BenedekFarkas:
> > @BenedekFarkas Question: Shouldn't workflow/action references throughout the whole repo be updated? [See this commit](https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/commit/28ce10657a408386353082c78bf24966e1ac3078) that was triggered by the release branch and created the tag.
> 
> Yes, the reference should be updated in the entire repository. Let me try and replicate this by forking the repo and trying to create a release in my own remote.
> 
> I'll have to check but I think this might be because it's running in a fork. But it may no longer be an issue now that the action inputs were surfaced up to the calling workflows.

OK, thanks, I'll run some functional tests too on a repo that isn't fork, just a mirror.

.github/workflows/tag-version-this-repo.yml
BenedekFarkas: I've made some updates, @Piedone you can check.

Piedone: That nicely describes what happens and how, but not why. So I'd change two things:

- Add something like this as the first paragraph under the section: "To release versions of Lombiq GitHub Actions, and allow consumers to reference both reusable workflows and composite actions in a versioned way (i.e. `@v1.0`) we employ some automation.
- Make the list a numbered one, since it describes steps in order.

.github/workflows/tag-version.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Commit and Push Changes
+        uses: stefanzweifel/git-auto-commit-action@8756aa072ef5b4a080af5dc8fef36c5d586e521d # v5
+        with:
+          commit_message: 'Set GitHub Actions/Workflows to tag ${{ steps.determine-tag.outputs.tagname }}'
+          tagging_message: ${{ steps.determine-tag.outputs.tagname }}
+
+      - name: Force Push Tag
+        continue-on-error: true
+        run: git push --tags --force
```
Piedone, line 101: Also publish a release on GitHub, please. We do this [during NuGet publishing too](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/publish-nuget/action.yml#L216). Basically this should suffice:

```yaml
    - name: Create Release
      # v1.11.2
      uses: ncipollo/release-action@18eadf9c9b0f226f47f164f5373c6a44f0aae169
      # This is to prevent creating releases when pushing tags for issue-specific pre-releases like
      # v4.3.1-alpha.osoe-86.
      if: "!contains(steps.setup.outputs.publish-version, '-')"
      with:
        allowUpdates: true
        generateReleaseNotes: true
```

However, at this point, it would be worth factoring it out to a central internally used action, so it's easier to keep its version updated all at once (instead of these two places), much like how I did it here: https://github.com/Lombiq/GitHub-Actions/pull/336/files.

davidpuplava: Ok, I can add that in and look to centralize that into it's own internal action.

Piedone: Thanks!

.github/workflows/tag-version-this-repo.yml
davidpuplava: Thanks @Piedone and @BenedekFarkas, I'll make these README edits and push shortly.

.github/workflows/tag-version.yml
davidpuplava: Refactored to internal action `release-action` in e0ce7b52744b17f81bbd90f7c60ea9b9afc43547.

EDIT: Updated `publish-nuget` action to use it and added it to `tag-version`.

EDIT again: fixed incorrect GitHub Action expression syntax in 1667cd22bd44b87e7e3d5cab95b6425c82621ffe.

.github/workflows/tag-version.yml
davidpuplava: Fixed in 36bbcefab51d7f5a744a026b06f1a30f685f6bf5.

.github/workflows/tag-version.yml
davidpuplava: Fixed in 36bbcefab51d7f5a744a026b06f1a30f685f6bf5.

davidpuplava:
Ok, I believe I now have this working. I successfully created the release in my forked repo: https://github.com/davidpuplava/GitHub-Actions/releases/tag/v0.1.

I had to elevate the GITHUB_TOKEN to give it `contents: write` permission, but I don't think that is necessary in LGHA based on what I learned from past issues.

.github/workflows/tag-version-this-repo.yml
BenedekFarkas: > And because the workflow is triggered on the push event to any release/** branch, it kicks off a 2nd workflow run.

Yep, that makes sense the we need to prevent it from re-triggering itself.

> One problem I see is any additional commits pushed to the release/v1.0 branch, fixing a small bug for example, may be challenging.

That's fine, if a subsequent fix need to be released after it, then that will be in a PR that targets the default branch and then we can start the release process from that PR's merge commit. The release branches can remain "floating", Microsoft does the same in the https://github.com/actions/runner-images repo for example.

-----

BTW my initial thought was that we should try to avoid wiring that condition in hoping that the process can be re-used, for example if we'd publish actions on the marketplace (that would require the action/workflow to exist separately in a repo) or even outside Lombiq, but at the moment it looks that would be a lot of additional complexity.

.github/actions/release-action/action.yml
BenedekFarkas, line 1: That's a lot of parameters! :)
Usually when we implement a wrapper action we don't include all the possible parameters if the default values are OK and we don't expect the need to override them, see [checkout](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/checkout/action.yml) for example (although it's a bit of an extreme). We can remove most of these so there's less of a chance of a breaking change when we upgrade the wrapped action to a new version.

BenedekFarkas (changes requested):

.github/actions/release-action/action.yml
davidpuplava: Yes, it took a little time to try and format the descriptions to wrap around :)

Understood. I'll remove all the parameters except for what we use in the `publish-nuget` and `tag-version` actions. It makes sense to limit the parameters and only add them as we go along.

BenedekFarkas: Yeah, sorry about the hassle, but I believe it will be easier to maintain this way.

BenedekFarkas:
> Ok, I believe I now have this working. I successfully created the release in my forked repo: https://github.com/davidpuplava/GitHub-Actions/releases/tag/v0.1.
> 
> I had to elevate the GITHUB_TOKEN to give it `contents: write` permission, but I don't think that is necessary in LGHA based on what I learned from past issues.

Are the other references updating for you (= in all actions/workflows pointing to Lombiq GitHub Actions stuff)? I see the spelling dictionary prefix updating for me too in my test repo, but none of the other stuff: https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/commit/ae2421a415d28c8241b58587265d0cde6c496312

.github/actions/release-action/action.yml
davidpuplava: No problem, deleting the extra inputs was quick. Fixed in 24f5ce672593779dbe565f7f5b41cd2a704802b7.

davidpuplava:
> Are the other references updating for you (= in all actions/workflows pointing to Lombiq GitHub Actions stuff)? I see the spelling dictionary prefix updating for me too in my test repo, but none of the other stuff: https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/commit/ae2421a415d28c8241b58587265d0cde6c496312

No, they were not updating for me. The only reference that was updating was the spelling dictionary reference. But I did figure out what was going on.

The [Tag Version](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version.yml) workflow uses a [default value](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version.yml#L38) of the current repository's base URL as the "target" string to find called Workflows and Actions. For LGHA, this is `Lombiq/GitHub-Actions`, but for your forked repository it is `BenedekFarkas/GitHub-Actions-OSOE-735-Test` and for my forked repository it is `davidpuplava/GitHubActions`.

So in our forked repositories, nothing was matched because the [Set GitHub Actions References](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/actions/set-gha-refs/action.yml) action's Set-Called-GHA-Refs.ps1 script is [matching using that base URL include list](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/actions/set-gha-refs/Set-Called-GHA-Refs.ps1#L15) (i.e. looking for `"uses: Lombiq/GitHub-Actions/..."` in LGHA, but looking for `"uses: BenedekFarkas/GitHub-Actions-OSOE-735-Test/..."` in your fork).

To see this in action, in your repository, you can modify the [Tag Version (this repo)](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version-this-repo.yml) workflow to use the [called-repo-base-include-list](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version.yml#L36) input and pass in `Lombiq/GitHub-Actions` so that all the references get updated.

See an example of this in my forked copy [here](https://github.com/davidpuplava/GitHub-Actions/blob/release/v0.1/.github/workflows/tag-version-this-repo.yml#L16). I included both my repo base and LGHA repo base, and it updated all the corresponding GitHub Action calls accordingly in the [automatic commit](https://github.com/davidpuplava/GitHub-Actions/commit/d9a35ad2e19feea8fcccab057658d8e7ac78b511).

During testing in my forked copy, I noticed a bug with the regex used to get the Git reference, which I fixed in 05a109a0031c0c81942965fcaa86e8511945fff8.

I also fixed a typo in getting the line number for the matched/changed file, and I fixed in 4d5e3a86532867fd21c67d7a7d2f2b0eaf6fc904.

BenedekFarkas:
> To see this in action, in your repository, you can modify the [Tag Version (this repo)](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version-this-repo.yml) workflow to use the [called-repo-base-include-list](https://github.com/Lombiq/GitHub-Actions/blob/issue/OSOE-735/.github/workflows/tag-version.yml#L36) input and pass in Lombiq/GitHub-Actions so that all the references get updated.

Gotcha, [that worked for me too](https://github.com/BenedekFarkas/GitHub-Actions-OSOE-735-Test/actions/runs/8300385160), so it was PIBKAC! :D

What did you to make the `github-actions[bot]` user commit instead of your own?

BTW one thing I noticed is that specifying multiple repositories via the `called-repo-base-include-list` will upgrade them to the same version. That's intentional and "obvious", just calling this out because we can still use synchronized versioning for actions/workflows across different repositories, but in that case we have to make sure that a new version number is released across each affected repository.

@Piedone please check and resolve the threads above you started, if they're GTG.

BenedekFarkas (approved):

Piedone:
I checked those two in the code, looks good!

Piedone:
Shouldn't this be merged now?

BenedekFarkas:
Yep, we're good to go - sorry @davidpuplava for the delay.

BenedekFarkas:
And here's the first release, thanks @davidpuplava! https://github.com/Lombiq/GitHub-Actions/releases/tag/v1.0

PR #343, opened 2024-04-29
OSOE-842: Fixing NuGetBuild csproj conditions during NuGet publishing 

### LINKED ISSUE #342: Fix NuGetBuild not taking effect in the publish-nuget action (OSOE-842)
See: https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/8836448057/job/24264925167

It seems like [-p:NuGetBuild=true](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/publish-nuget/action.yml#L189) is not applied.

This is indicated here, because the action searches for `Lombiq.HelpfulLibraries.OrchardCore.csproj` instead of using the NuGet package: https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/8836448057/job/24264925167#step:4:207 and also here for `Lombiq.Npm.Targets.props`: https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/8836448057/job/24264925167#step:4:1304

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-842)

Piedone:
[OSOE-842](https://lombiq.atlassian.net/browse/OSOE-842)
Fixes #342

[OSOE-842]: https://lombiq.atlassian.net/browse/OSOE-842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
+++ b/.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
@@ -41,10 +41,11 @@ foreach ($projectFile in $projectFiles)
     # Save the changes back to the .csproj file.
     $projectXml.Save($projectFile)
 
-    # --no-restore because we run it in a separate step with a configured output verbosity, what's not possible with
-    # the package command (and letting it implicitly restore here also makes NuGet publishing take 2-3 times as much).
-    # Unfortunately, this also makes it skip compatibility checks.
-    dotnet add $projectFile.FullName package 'Microsoft.SourceLink.GitHub' --source 'https://api.nuget.org/v3/index.json' --no-restore
+    # We can't use --no-restore because not only would it skip checks, it'd also be incompatible with projects using
+    # Central Package Management (https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management).
+    # Unfortunately, this makes NuGet publishing a lot slower than using dotnet restore, and it's also more verbose
+    # (without the ability to configure that verbosity).
+    dotnet add $projectFile.FullName package 'Microsoft.SourceLink.GitHub' --source 'https://api.nuget.org/v3/index.json'
 }
 
 Write-Output 'SourceLink package added to all projects.'
--- a/.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
+++ b/.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
@@ -19,6 +19,8 @@ foreach ($projectFile in $projectFiles)
 {
     # Below we first prepare the project file by adding the NuGetBuild=true property to the top of it. This is needed
     # for dotnet add package which could otherwise fail due to conditions in the project file.
+    # We don't use a Directory.Build.props file for this because the project might have one already, and then we'd need
+    # to merge it anyway.
 
     # Load the project file as XML.
     $projectXml = [xml](Get-Content $projectFile)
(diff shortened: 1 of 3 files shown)


.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
```diff
@@ -17,5 +17,32 @@ else
...
+    # Add the new PropertyGroup to the project as the first node, to make sure that it's fully applied.
+    $projectXml.Project.InsertBefore($propertyGroup, $projectXml.Project.FirstChild)
+
+    # Save the changes back to the .csproj file.
+    $projectXml.Save($projectFile)
+
+    # --no-restore because we run it in a separate step with a configured output verbosity, what's not possible with
+    # the package command (and letting it implicitly restore here also makes NuGet publishing take 2-3 times as much).
+    # Unfortunately, this also makes it skip compatibility checks.
+    dotnet add $projectFile.FullName package 'Microsoft.SourceLink.GitHub' --source 'https://api.nuget.org/v3/index.json' --no-restore
```
sarahelsaig: Running the command with `--no-restore` adds `<PackageReference Include="Microsoft.SourceLink.GitHub" Version="*" />` to the project. This breaks when you use _Directory.Packages.props_ like in OCC like this:

> error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: Microsoft.SourceLink.GitHub.

btw I don't understand the "because we run it in a separate step with a configured output verbosity, what's not possible with the package command". Could you please elaborate?

sarahelsaig (changes requested):

.github/actions/publish-nuget/Add-SourceLinkPackage.ps1
Piedone: I see, that's a pity.

With implicit restore, this command will do a package restore with a very verbose output (it looks more verbose than the default `normal` for `dotnet` commands, and definitely a lot more verbose than `quiet`, what's the default for our workflows). Most possibly it's partially this, that it slows NuGet package publishing down a lot (not just this step, but the whole thing will take 2-2,5 minutes instead of <1 minute for UITT, for example).

sarahelsaig: Yes, you can't configure the verbosity of `dotnet package add`, but I don't think that extra 10 lines slows things down. (otherwise you could always do `> $null` to discard the stdout) Rather it's because the CLI opens a remote connection to fetch the correct version and check for dependencies.
image

Piedone: It's more like 2000 extra lines, see e.g. [here](https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/8884157839/job/24436095344#step:4:112) (contrast this with e.g. [this other one](https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/8898351256/job/24435314424#step:4:110)). Discarding the whole output is not an option, because we need to know about that information in case something fails. I really don't want to go down the "filter the stdoutput for anything that's always useful" route, and it wouldn't help with the performance anyway.

But this is just FYI, it doesn't matter if we need to use it to work with OCC.

sarahelsaig: I will have to look into this, but I vaguely remember that slow console out is a Windows-specific issue while the publish-nuget runs on Ubuntu. Anyway, as you said not important for now.

sarahelsaig (approved):

PR #351, opened 2024-05-17
OSOE-744: Add wrapper for the markdownlint-cli2 action

### LINKED ISSUE #289: Add wrapper for the markdownlint-cli2 action (OSOE-744)
[markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2-action) looks like the shortest path to standalone Markdown linting that doesn't rely on [Lombiq.NodeJs.Extensions](https://github.com/Lombiq/NodeJs-Extensions) (which needs a solution to be built in).

The action is developed by the author of the [markdownlint](https://github.com/DavidAnson/markdownlint) library, which is what Lombiq.NodeJs.Extensions uses as well.

Useful for documentation-heavy repositories, such as this one, so we don't have to wait for the OSOCE build to surface MD linting errors, similarly how this repository has its own spell checking.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-744)

BenedekFarkas:
[OSOE-744](https://lombiq.atlassian.net/browse/OSOE-744)
Fixes #289

Note: The `markdown-lint` action and `lombiq.markdownlint.json` contains comments to warn about synchronizing the version of `markdownlint` and configuration changes to Lombiq.NodeJs.Extensions (and vice versa). This is a slight inconvenience (having to maintain this in two places), but these are completely independent libraries providing different means to execute linting and the associated configuration changes are extremely rare/unlikely.

[OSOE-744]: https://lombiq.atlassian.net/browse/OSOE-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/markdown-lint/lombiq.markdownlint.json
```diff
@@ -0,0 +1,10 @@
+// When updating, also update the same configuration file in Lombiq.NodeJs.Extensions.
+{
+    "default": true,
+    "MD004": { "style": "dash" },
+    "MD013": false,
+    "MD033": { "allowed_elements": [ "kbd" ] },
+    "MD049": { "style": "underscore" },
+    "MD050": { "style": "asterisk" }
+  }
+
```
sarahelsaig: Some leftover spaces.
```suggestion

```

.github/actions/markdown-lint/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    description: String to use as a separator for the "globs" input (defaults to newline).
+    default: "\n"
+    required: false
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Markdown Linting
+    # When upgrading, also update Lombiq.NodeJs.Extensions to use the corresponding version of the markdownlint package.
+      uses: DavidAnson/markdownlint-cli2-action@b4c9feab76d8025d1e83c653fa3990936df0e6c8 # v16.0.0
```
sarahelsaig: imo the comment should have the same indentation as the `uses`.
```suggestion
    - name: Markdown Linting
      # When upgrading, also update Lombiq.NodeJs.Extensions to use the corresponding version of the markdownlint package.
      uses: DavidAnson/markdownlint-cli2-action@b4c9feab76d8025d1e83c653fa3990936df0e6c8 # v16.0.0
```

.github/actions/markdown-lint/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    default: ''
+    required: false
+  globs:
+    description: Glob expression(s) of files to lint (newline-delimited).
+    default: '**/*.{md,markdown}'
+    required: false
+  separator:
+    description: String to use as a separator for the "globs" input (defaults to newline).
+    default: "\n"
+    required: false
```
sarahelsaig, line 24: Since these are the same inputs as the [the called action](https://github.com/DavidAnson/markdownlint-cli2-action/blob/main/action.yml), please include a comment with the link to make future maintenance easier.

.github/actions/markdown-lint/action.yml
```diff
@@ -0,0 +1,35 @@
...
+# When updating defaults here also update them in the `markdown-lint` workflow.
+inputs:
+  config:
+    description: Path to a file to use for the base configuration object (defaults to "lombiq.markdownlint.json").
+    default: 'lombiq.markdownlint.json'
+    required: false
+  fix:
+    description: Whether to fix issues automatically for the rules that support it (any truthy value enables).
+    default: ''
+    required: false
```
sarahelsaig, line 16: Have you tested this feature? How does this even work? Maybe it would be cool to somehow copy the auto-fixed files into an artifact if that's possible.

.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,73 @@
...
+    name: Markdown Linting
+    runs-on: ubuntu-22.04
+    timeout-minutes: ${{ inputs.timeout-minutes }}
+    steps:
+      - name: Checkout
+        uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+        with:
+          token: ${{ secrets.CHECKOUT_TOKEN }}
+
+      # This is a workaround to be able to check submodules too in the repository.
```
sarahelsaig: Could you elaborate what's the problem that prevents from checking submodules?

.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Stub repo layout
+        shell: pwsh
+        run: |
+          git config --global user.email "you@example.com"
+          git config --global user.name "Your Name"
+          Remove-Item .\.git\ -recurse -force
```
sarahelsaig: Powershell switches should be capitalized.
```suggestion
          Remove-Item .\.git\ -Recurse -Force
```

sarahelsaig (changes requested):

.github/actions/markdown-lint/action.yml
BenedekFarkas: Yep, thanks!

.github/workflows/markdown-lint.yml
BenedekFarkas: TBH I don't know, I started this workflow as a copy of `spelling`, but `yaml-lint` doesn't have this step. I'll remove it for now and do a test run in OSOCE to see if it works without it.

.github/workflows/markdown-lint.yml
BenedekFarkas: I'l fix this in the spelling workflow instead.

.github/actions/markdown-lint/action.yml
BenedekFarkas: There's a link in the description above.

.github/actions/markdown-lint/action.yml
BenedekFarkas: I haven't, but I'll check!

.github/actions/markdown-lint/action.yml
sarahelsaig: That's a link to the repo, you'd still have to find the config files in it. Also I don't think the comment using the term "wrapping" is sufficiently clear that all these parameters are copied from the action.

.github/workflows/markdown-lint.yml
sarahelsaig: Yeah, I see it was added 2 years ago when .github/workflows/spelling.yml was originally created. If the problem is that the submodules aren't initialized by default (so just empty directories) then the correct way to do this is a simple one-liner: `git submodule update --init`.

.github/actions/markdown-lint/action.yml
BenedekFarkas: I've implemented this, see https://github.com/Lombiq/GitHub-Actions/actions/runs/9354126597 for the logs and artifact. [The next commit](https://github.com/Lombiq/GitHub-Actions/pull/351/commits/7d0dca2eb9e89a8ba456f92447c5e0213e24d26c) contains those fixes.

.github/actions/markdown-lint/action.yml
```diff
@@ -0,0 +1,70 @@
...
+      shell: pwsh
+      run: (Resolve-Path "${{ github.action_path }}/../../../Scripts").Path >> $Env:GITHUB_PATH
+
+    - name: Copy files updated by markdown-lint automatic fixes
+      if: inputs.fix == 'true'
+      id: markdown-lint-fix-files
+      shell: pwsh
+      run: |
+        $artifactFolder = [System.Guid]::NewGuid()
+        $files = git diff --name-only '*.md' '*.markdown'
```
BenedekFarkas, line 49: This filter is static (as in wired in to include `.md` and `.markdown` files), unlike the `globs` parameter, but IMO this is good enough.

.github/actions/markdown-lint/action.yml
BenedekFarkas: Gotcha, here it is: 095986d1158cd861db460cc61f4b7a8f2bdb0d6a

sarahelsaig (approved):

PR #363, opened 2024-06-12
OSOE-580: Custom analyzer rule for parameter and variable name styling

BenedekFarkas:
[OSOE-580](https://lombiq.atlassian.net/browse/OSOE-580)

[OSOE-580]: https://lombiq.atlassian.net/browse/OSOE-580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/test-dotnet/Merge-FailureDumps.ps1
+++ b/.github/actions/test-dotnet/Merge-FailureDumps.ps1
@@ -8,4 +8,4 @@ $rootDirectory = (Test-Path -Path $testDirectory) ? $testDirectory : $Directory
 Get-ChildItem $rootDirectory -Recurse |
     Where-Object { $PSItem.Name -eq 'FailureDumps' } |
     ForEach-Object { $PSItem.GetDirectories() } |
-    ForEach-Object { Move-Item $PSItem.FullName "$Directory/FailureDumps/${_.Name}" }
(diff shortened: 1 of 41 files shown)


.github/actions/test-dotnet/Merge-FailureDumps.ps1
```diff
@@ -8,4 +8,4 @@ $rootDirectory = (Test-Path -Path $testDirectory) ? $testDirectory : $Directory
 Get-ChildItem $rootDirectory -Recurse |
     Where-Object { $PSItem.Name -eq 'FailureDumps' } |
     ForEach-Object { $PSItem.GetDirectories() } |
-    ForEach-Object { Move-Item $PSItem.FullName "$Directory/FailureDumps/${_.Name}" }
+    ForEach-Object { Move-Item $PSItem.FullName "$Directory/FailureDumps/$($PSItem.Name)" }
```
BenedekFarkas, line 11: FYI this is a functional change, not just cosmetic. Variable names enclosed in braces are used to be able to use special characters in the variable name, so it won't behave the same as `$_.Name` (or `$PSItem.Name`). I also removed usages of braced variables in other files.

Idea: Add an analyzer that detects that braced variables are used unnecessarily, which is basically when you don't have a special character enclosed in the braces.

sarahelsaig (approved):

PR #364, opened 2024-06-14
OSOE-855: Not setting the dotnet-quality config for concrete .NET versions

Piedone:
[OSOE-855](https://lombiq.atlassian.net/browse/OSOE-855)

[OSOE-855]: https://lombiq.atlassian.net/browse/OSOE-855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-dotnet/action.yml
+++ b/.github/actions/setup-dotnet/action.yml
@@ -18,15 +18,8 @@ runs:
     - name: Set Environment Variables
       shell: pwsh
       run: |
-        # This ceremony is needed to make the env vars available in subsequent steps too, see:
-        # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable.
-        # Note that in PowerShell, echo is not needed. Make sure not to put spaces around the equal signs when writing
-        # to $Env:GITHUB_ENV.
-        "DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true" >> $Env:GITHUB_ENV
-        $Env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = "true"
-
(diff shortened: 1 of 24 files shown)


.github/actions/setup-dotnet/action.yml
```diff
@@ -23,6 +28,17 @@ runs:
...
 
+    # For wildcard .NET versions, the General Availability ("ga") channel should be selected. For concrete versions,
+    # this shouldn't be set, because that'll cause a warning.
+    - name: Set dotnet-quality
+      id: set-dotnet-quality
+      shell: pwsh
+      run: |
+        if ('${{ inputs.dotnet-version }}' -match 'x')
+        {
+            Set-GitHubOutput "dotnet-quality" "ga"
```
BenedekFarkas: Since you've set up the scripts to be able to use `Set-GitHubOutput`, it would be better to update the rest of the steps to use it as well.
BTW you can just use singlequotes here (the strings are constant).

BenedekFarkas (changes requested):

.github/actions/setup-dotnet/action.yml
Piedone: The other steps don't write to the output, but `GITHUB_ENV`, and we don't have a script for that. But now I created it and changed the references.

BenedekFarkas: Oh, yeah, I mixed them up! But it's a nice improvement in any case.

BenedekFarkas (approved):

.github/actions/setup-dotnet/action.yml
Piedone: Yep!

PR #376, opened 2024-07-20
OFFI-56: Adding Elasticsearch setup workflow

porgabi:
[OFFI-56](https://lombiq.atlassian.net/browse/OFFI-56)

[OFFI-56]: https://lombiq.atlassian.net/browse/OFFI-56?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wAsnk (approved):

.github/actions/setup-elasticsearch/action.yml
```diff
@@ -4,14 +4,6 @@ description: Sets up Elasticsearch
 runs:
   using: composite
   steps:
-    - name: Configure sysctl limits
```
wAsnk: Why don't we need this, while the documentation says otherwise? https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch

porgabi: As far as I can tell, this is already included in the action that's called: https://github.com/elastic/elastic-github-actions/blob/master/.github/workflows/elasticsearch.yml

wAsnk: But you are using this https://github.com/elastic/elastic-github-actions/blob/master/elasticsearch/action.yml action, not the workflow you linked, which is totally fine and it should be like that.

porgabi: Hmm. It's odd that it works this way too. Either way, I'll restore the sysctl limits step.

wAsnk (approved):

Piedone:
There's a container build step running before anything else, even if Elasticsearch setup is not enabled:

image

See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/10082467596/job/27876788040.

Can this be avoided?

porgabi:
> There's a container build step running before anything else, even if Elasticsearch setup is not enabled:
> 
> image
> 
> See https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/10082467596/job/27876788040.
> 
> Can this be avoided?

I can't quite tell why referencing the third-party action without actually running it does this.

Piedone:
Then you can look up if this is normal behavior for such an action using Docker and/or ask under the project.

PR #378, opened 2024-07-25
OCC-245: Add the option to publish without Microsoft.SourceLink.GitHub

sarahelsaig:
[OCC-245](https://lombiq.atlassian.net/browse/OCC-245)
This intentionally targets `dev`, as it's not specific to OC 2.0.

Package publishing to Cloudsmith fails if Microsoft.SourceLink.GitHub was added to the project.

[OCC-245]: https://lombiq.atlassian.net/browse/OCC-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -35,6 +35,10 @@ inputs:
       https://github.com/actions/upload-artifact#retention-period for more details. Note that this only affects the
       retention of the workflow run's artifact, not the artifacts attached to the release created on GitHub; those will
       remain indefinitely.
+  add-source-link-package:
+    required: false
+    default: 'true'
+    description: If set to "true", the Microsoft.SourceLink.GitHub NuGet package is added to the projects.
 
 runs:
   using: 'composite'
@@ -113,7 +117,15 @@ runs:
     # This also implicitly restores NuGet dependencies.
     - name: Add Source Link package
       shell: pwsh
-      run: dotnet restore
+      run: |
+        if ('${{ inputs.add-source-link-package }}' -eq 'true')
+        {
+          Add-SourceLinkPackage
+        }
+        else
+        {
+          dotnet restore
+        }
 
     - name: Update package manifest version
       shell: pwsh
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -121,11 +121,11 @@ runs:
       run: |
         if ('${{ inputs.add-source-link-package }}' -eq 'true')
         {
-          Add-SourceLinkPackage
+            Add-SourceLinkPackage
         }
         else
         {
-          dotnet restore -p:NuGetBuild=true
+            dotnet restore -p:NuGetBuild=true
         }
 
     - name: Update package manifest version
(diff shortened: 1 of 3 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -114,7 +118,15 @@ runs:
...
-      run: Add-SourceLinkPackage
+      run: |
+        if ('${{ inputs.add-source-link-package }}' -eq 'true')
+        {
+          Add-SourceLinkPackage
+        }
+        else
+        {
+          dotnet restore -p:NuGetBuild=true
+        }
```
Piedone:
```suggestion
        if ('${{ inputs.add-source-link-package }}' -eq 'true')
        {
            Add-SourceLinkPackage
        }
        else
        {
            dotnet restore -p:NuGetBuild=true
        }
```

.github/actions/publish-nuget/action.yml
```diff
@@ -114,7 +118,15 @@ runs:
     # This also implicitly restores NuGet dependencies.
     - name: Add Source Link package
       shell: pwsh
-      run: Add-SourceLinkPackage
+      run: |
+        if ('${{ inputs.add-source-link-package }}' -eq 'true')
+        {
+          Add-SourceLinkPackage
```
Piedone: How does this fail with Cloudsmith?

Piedone (changes requested):

.github/actions/publish-nuget/action.yml
sarahelsaig: Like this: https://github.com/Lombiq/Helpful-Libraries/actions/runs/9643961750/job/26595027509#step:4:288

> error: Package 'Microsoft.SourceLink.GitHub' is incompatible with 'all' frameworks in project '/home/runner/work/Helpful-Libraries/Helpful-Libraries/Lombiq.HelpfulLibraries.LinqToDb/Lombiq.HelpfulLibraries.LinqToDb.csproj'.

(All the other "Unable to resolve" errors in the log are irrelevant and they also went away after I turned off `Microsoft.SourceLink.GitHub`.)

Piedone: So it tries to restore all packages, as we observed earlier (hence the code in the script) but it only uses nuget.org and thus fails. Isn't this fixed by removing `--source` [here](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/publish-nuget/Add-SourceLinkPackage.ps1#L48)? Only if this still works if there's no NuGet.config.

sarahelsaig: Perhaps. Do we know why was `--source 'https://api.nuget.org/v3/index.json'` necessary in the first place?

sarahelsaig: > Only if this still works if there's no NuGet.config.

Yes, that works without `--source`:
image

Piedone: Great, then we can just use that.

I think when writing that my understanding was that `--source` will use that source to restore SourceLink itself. If it works without it, then nothing else to do.

Piedone (approved):

PR #394, opened 2024-10-25
OSOE-864: Automatically check baseline package validation during review and update check-spelling to 0.0.24

### LINKED ISSUE #359: Automatically check baseline package validation during review (OSOE-864)
Right now it's too easy to accidentally introduce breaking changes, why not run the same check we do during `publish-nuget` also during the normal review flow? I suggest adding a label like `validate-baseline` which would run the check and write a comment on the PR if there are any results.

The same logic should be run by `done-jira-issue-if-checks-succeed` as well. This way the check would run once before the issues goes to review, giving useful info both the contributor and the reviewer.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-864)

Piedone:
[OSOE-864](https://lombiq.atlassian.net/browse/OSOE-864)
Fixes #359

[OSOE-864]: https://lombiq.atlassian.net/browse/OSOE-864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build-and-test-dotnet.yml
```diff
@@ -153,7 +153,7 @@ on:
       timeout-minutes:
         type: number
         default: 360
-        description: Configuration for the timeout-minutes parameter of the workflow. The 360 is GitHub's default.
+        description: Configuration for the timeout-minutes parameter of the workflow. 360 is GitHub's default.
```
Piedone, line 156: The "the" in these referred to the default, but it's bad English, and in some occasions, the default wasn't even 360.

.github/workflows/publish-nuget.yml
```diff
@@ -39,7 +38,7 @@ on:
       timeout-minutes:
         type: number
         default: 15
-        description: Configuration for the timeout-minutes parameter of the workflow. The 360 is GitHub's default.
+        description: Configuration for the timeout-minutes parameter of the workflow. 360 is GitHub's default.
```
sarahelsaig, line 41: It says "360 is GitHub's default" but [the actual default is 15](https://github.com/Lombiq/GitHub-Actions/blob/7780a0a775c0a101bd00bd27cce45b5b1eb6b56c/.github/workflows/publish-nuget.yml#L40). Should it be this instead?

```suggestion
        description: Configuration for the timeout-minutes parameter of the workflow. 15 is GitHub's default.
```

.github/actions/publish-nuget/action.yml
```diff
@@ -39,6 +42,12 @@ inputs:
     required: false
     default: 'true'
     description: If set to "true", the Microsoft.SourceLink.GitHub NuGet package is added to the projects.
+  dry-run:
+    required: false
+    default: 'false'
+    description: |
+      If set to "true", the action will perform all package creation and validation steps without actually publishing to
+      NuGet or creating a GitHub release.
```
sarahelsaig: It's always a NuGet package, but you don't necessarily publish _to_ NuGet.
```suggestion
      If set to "true", the action will perform all package creation and validation steps without actually publishing
      the NuGet package or creating a GitHub release.
```

.github/workflows/publish-nuget.yml
```diff
@@ -69,6 +68,12 @@ on:
         type: string
         default: 'true'
         description: If set to "true", the Microsoft.SourceLink.GitHub NuGet package is added to the projects.
+      dry-run:
+        type: string
+        default: 'false'
+        description: |
+          If set to "true", the action will perform all package creation and validation steps without actually publishing
+          to NuGet or creating a GitHub release.
```
sarahelsaig: Same here.
```suggestion
          the NuGet package or creating a GitHub release.
```

Docs/Workflows/Productivity/PublishNuGetPackage.md
```diff
@@ -15,14 +15,14 @@ jobs:
...
+      API_KEY: ${{ secrets.DEFAULT_NUGET_PUBLISH_API_KEY }}
 ```
 
 The _publish.yml_ workflow is triggered on a tag pushed to any branch with the prefix `v` and should contain a [valid version number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics), e.g. `v1.0.1`, which will be extracted and used to version the NuGet packages produced.
 
 There is no configuration required for automated Orchard Core extension manifest versioning, all of the `Manifest.cs` files are looked up, and the existing `Version` properties are updated automatically inside the `Module` or `Theme` definition with the version pushed. Note that the `Version` property should be present and leave the version number on the default value (0.0.1). This is because we don't actually need to keep manifest version changes in the code.
 
-It takes one non-optional secret parameter, `apikey`, the organization API key for pushing to NuGet, and two optional parameters, `source` and `verbosity`. E.g.:
+It takes one non-optional secret parameter, `API_KEY`, the organization API key for pushing to NuGet, and two optional parameters, `source` and `verbosity`. E.g.:
```
sarahelsaig, line 25: This is not a breaking change, just the documentation was out of date, right?

sarahelsaig (changes requested):

.github/workflows/publish-nuget.yml
Piedone: This is about the default timeout for every GitHub workflow, which is 360. This is added as information for context.

Docs/Workflows/Productivity/PublishNuGetPackage.md
Piedone: Yep, `apikey` was the name of it only in ancient times.

.github/workflows/publish-nuget.yml
sarahelsaig: It's still misleadingly phrased. In that case you'd want to write
```suggestion
        description: Configuration for the timeout-minutes parameter of the workflow. GitHub's default timeout is 360.
```
but I don't think it merits mentioning because it's irrelevant for this specific input (whose default value is 15 and not 360 like with the other `timeout-minutes`s above).

Piedone: Yeah, fair enough.

sarahelsaig (approved):

PR #400, opened 2024-11-10
OSOE-402: Disabling package validation for prereleases during NuGet publishing

Piedone:
[OSOE-402](https://lombiq.atlassian.net/browse/OSOE-402)

[OSOE-402]: https://lombiq.atlassian.net/browse/OSOE-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -199,10 +199,13 @@ runs:
         ${{ inputs.dotnet-pack-ignore-warning }}
         '@.Split() | ? { $_ }
 
+        $version = '${{ steps.setup.outputs.publish-version }}'
+        $enablePackageValidation = !($version -like '*-*')
+
         $parameters = @{
-          EnablePackageValidation=$true
+          EnablePackageValidation=$enablePackageValidation
           PackageValidationBaselineVersion='${{ steps.setup.outputs.baseline-version }}'
-          Version='${{ steps.setup.outputs.publish-version }}'
+          Version=$version
           PackParameters = @(
             "--configuration:Release",
             "--warnaserror",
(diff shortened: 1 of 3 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -199,10 +199,13 @@ runs:
         ${{ inputs.dotnet-pack-ignore-warning }}
         '@.Split() | ? { $_ }
 
+        $version = '${{ steps.setup.outputs.publish-version }}'
+        $enablePackageValidation = !($version -like '*-*')
```
dministro, line 207: My impression is, this conditional package validation can easily lead to confusion.
I know changing the workflow parameter on the issue branch before publishing a prerelease version does not improve the developer experience, but please consider exposing this as a script parameter.

Piedone: I assume you mean action (and workflow?) parameter, since it's already a script parameter. How would that work, though? Would it override this alpha detection logic (we can have the default for it as empty string to allow this)?

We need this logic in any case, since not only has this been the behavior for a very long time, but it's also useful, since we never need package validation for alphas.

dministro: > ..., but it's also useful, since we never need package validation for alphas.

If it's really not necessary, then it's okay. Previously, it was always validated, which is why I have a bad feeling here.

Piedone: Hmm, wait, indeed it was always validated, I've mistaken it with the other exclusions for alphas. However, still, I think it should always be disabled for alphas. This is due to errors like this:

https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/11769438842/job/32780456521

Here, the suppression file was added to satisfy Validate NuGet Publish (what [passed for the same commit])https://github.com/Lombiq/UI-Testing-Toolbox/actions/runs/11768091624), and rightfully so, since the same will be needed for a non-alpha NuGet publish too. However, that clashes with package validation for alphas. It seems that package validation not only doesn't require a _CompatibilitySuppressions.xml_ for breaking changes when publishing alpha versions but actively disallows it.

I added a comment about this.

dministro (approved):

PR #403, opened 2024-12-11
OSOE-928: Set DOTNET_INSTALL_DIR NUGET_PACKAGES to D Drive on Windows

### LINKED ISSUE #402: Check if moving the .NET SDK install dir to the D drive on Windows makes the builds faster (OSOE-928)
See https://github.com/actions/setup-dotnet/issues/260#issuecomment-2516145018.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-928)

Piedone:
[OSOE-928](https://lombiq.atlassian.net/browse/OSOE-928)
Fixes #402

[OSOE-928]: https://lombiq.atlassian.net/browse/OSOE-928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-dotnet/action.yml
+++ b/.github/actions/setup-dotnet/action.yml
@@ -24,7 +24,7 @@ runs:
     # This makes builds faster under Windows, see
     # https://github.com/actions/setup-dotnet/issues/260#issuecomment-2516145018. Only GitHub-hosted standard (not
     # larger) runners have such a D drive.
-    - name: Set DOTNET_INSTALL_DIR NUGET_PACKAGES to D Drive on Windows
+    - name: Set DOTNET_INSTALL_DIR and NUGET_PACKAGES to D Drive on Windows
       if: runner.os == 'Windows' && !contains(runner.name, 'core_') && !contains(runner.name, 'warp')
       shell: pwsh
       run: |
--- a/.github/actions/setup-dotnet/action.yml
+++ b/.github/actions/setup-dotnet/action.yml
@@ -24,7 +24,7 @@ runs:
     # This makes builds faster under Windows, see
     # https://github.com/actions/setup-dotnet/issues/260#issuecomment-2516145018. Only GitHub-hosted standard (not
     # larger) runners have such a D drive.
-    - name: Set DOTNET_INSTALL_DIR on Windows
+    - name: Set DOTNET_INSTALL_DIR NUGET_PACKAGES to D Drive on Windows
       if: runner.os == 'Windows' && !contains(runner.name, 'core_') && !contains(runner.name, 'warp')
       shell: pwsh
       run: |
--- a/.github/actions/setup-dotnet/action.yml
+++ b/.github/actions/setup-dotnet/action.yml
@@ -29,6 +29,7 @@ runs:
       shell: pwsh
       run: |
         Set-GitHubEnv 'DOTNET_INSTALL_DIR' 'D:\dotnet'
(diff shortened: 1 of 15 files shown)


.github/actions/setup-dotnet/action.yml
```diff
@@ -21,6 +21,16 @@ runs:
         Set-GitHubEnv 'DOTNET_SKIP_FIRST_TIME_EXPERIENCE' 'true'
         Set-GitHubEnv 'DOTNET_CLI_TELEMETRY_OPTOUT' 'true'
 
+    # This makes builds faster under Windows, see
+    # https://github.com/actions/setup-dotnet/issues/260#issuecomment-2516145018. Only GitHub-hosted standard (not
+    # larger) runners have such a D drive.
+    - name: Set DOTNET_INSTALL_DIR NUGET_PACKAGES to D Drive on Windows
+      if: runner.os == 'Windows' && !contains(runner.name, 'core_') && !contains(runner.name, 'warp')
```
BenedekFarkas: I'd rather list (couldn't find a way to use a regex here) the exact runners this should be allowed to run on, because this expression would potentially let through 3rd-party runners other than WarpBuild or self-hosted ones.
```suggestion
      if: contains(fromJSON('["windows-latest", "windows-2019", "windows-2022"]'), runner.name)
```
We'll probably need to add `windows-2025` next year, but that should be tested first anyway.

.github/actions/setup-dotnet/action.yml
```diff
@@ -21,6 +21,16 @@ runs:
         Set-GitHubEnv 'DOTNET_SKIP_FIRST_TIME_EXPERIENCE' 'true'
         Set-GitHubEnv 'DOTNET_CLI_TELEMETRY_OPTOUT' 'true'
 
+    # This makes builds faster under Windows, see
+    # https://github.com/actions/setup-dotnet/issues/260#issuecomment-2516145018. Only GitHub-hosted standard (not
+    # larger) runners have such a D drive.
+    - name: Set DOTNET_INSTALL_DIR NUGET_PACKAGES to D Drive on Windows
```
BenedekFarkas:
```suggestion
    - name: Set DOTNET_INSTALL_DIR and NUGET_PACKAGES to D Drive on Windows
```

BenedekFarkas (changes requested):

.github/actions/setup-dotnet/action.yml
Piedone: `runner.name` contains a technical identifier of the runner instance, which is not the same as the machine type, what we supply to the workflow, but not this action, as `windows-latest and the likes. So, we can't check these here.

This is all the context that's available: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/12331366373/job/34417905198 and https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/12331388797/job/34417952382 I dislike checking what to _not_ match, because this is open to errors, but I don't see any other way: the standard runners don't have consistent names, but the larger ones do, and there's not much else information.

There's also `runner.environment` that can make this safer, what I added.

BenedekFarkas: Okey-dokey!

BenedekFarkas (approved):

PR #406, opened 2024-12-22
OSOE-935: Removing obsolete ui-test-parallelism, fixing blame hang dump merging

Piedone:
[OSOE-935](https://lombiq.atlassian.net/browse/OSOE-935)

[OSOE-935]: https://lombiq.atlassian.net/browse/OSOE-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/test-dotnet/Merge-BlameHangDumps.ps1
```diff
@@ -28,7 +28,7 @@ function ItemFilter($Item, $TestConfiguration)
     }
 
     $allow = (($Item.Name -like 'Sequence_*.xml') -or ($Item.Name -like '*_hangdump.dmp'))
-    if (-not ($allow -and $TestConfiguration))
+    if (-not $allow -and $TestConfiguration)
```
Piedone, line 31: I'm not sure about this, but this seems to be the original intent: `if NOT $allow and if $TestConfiguration is set`. This was the case before this change too: https://github.com/Lombiq/GitHub-Actions/commit/0fdf57fac05e5f1095e0383ea685803b1894cd55 (though the predecessor was syntactically incorrect).

.github/actions/test-dotnet/Merge-BlameHangDumps.ps1
```diff
@@ -55,5 +55,6 @@ Get-ChildItem $testDirectory.Path -Recurse |
             New-Item -Type Directory -Path $destinationDirectory
         }
 
-        Move-Item -Path $PSItem.FullName -Destination $destinationDirectory
+        # Moving the files would be faster, but the original files might only be readable by the current user.
+        Copy-Item -Path $PSItem.FullName -Destination $destinationDirectory
```
dministro, line 59: I used `Move-Item` here because - during the development - some runs ended with a `not enough free disk space` error. I don't know if this is still the case. Please consider to keep the `Move-Item` approach.

dministro (changes requested):

.github/actions/test-dotnet/Merge-BlameHangDumps.ps1
Piedone: I changed this because some ZAP files of the ZAP reports remained access denied, despite all my efforts (visible in the history, but I made sure that ZAP is absolutely shut down, also the whole of Docker, no other process is accessing the files, the permissions are correct). That broke blame hang dump collection in every single OSOCE run. I don't see any other solution than copying.

A blame hang dumps is less than a GB, BTW (see e.g. https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/12901077411), that should not be an issue even on a 2-core standard runner, let alone any larger ones (and all public projects on GitHub get the 4-core ones with more space).

dministro (approved):

PR #426, opened 2025-01-12
OSOE-960: Add codespell action and workflow for spell-checking

### LINKED ISSUE #423: Add codespell action and workflow for spell-checking (OSOE-960)
Add an action and a reusable workflow to do spell-checking with [codespell](https://github.com/codespell-project/codespell).

Codespell works by looking for misspellings ([like these](https://github.com/SimonVerhoeven/codespell/blob/master/codespell_lib/data/dictionary.txt)) instead of matching every word to a dictionary. While this is surely a less safe approach if the ultimate goal is to prevent spelling mistakes, it also looks like one that would basically eliminate false positives. This is a better deal than having frequent annoying false positives, even if this means that a few misspellings may slip through.
 
We could use [the codespell GHA action](https://github.com/codespell-project/actions-codespell), what installs codespell and runs it. However, this is only a quite thin wrapper on the codespell executable, doesn't use the latest version of it, and doesn't expose all config options (like https://github.com/codespell-project/actions-codespell/pull/19 and https://github.com/codespell-project/actions-codespell/issues/63) that we might need. So, perhaps just use it from PIP by doing something [like this](https://metaredux.com/posts/2021/11/26/automate-spell-checking-in-your-projects-with-github-actions.html).

- [x] Add an action and a reusable workflow. The action should run codespell from the CLI.
- [x] The action should contain a default configuration suitable for most of our projects, with options to override what's necessary. Using a custom [config file](https://github.com/codespell-project/codespell?tab=readme-ov-file#using-a-config-file) in the consumer repo should be possible.
- [x] It should be possible to exclude any number of complete files and folders, also by patterns (so we don't check all OSOCE submodules in other superprojects).
- [x] Display a hint/message on failing spell-checking with information about where to find documentation, and how to ignore single false positives inline.
- [x] Add Markdown documentation.
- [x] Add a job to the Build and Test workflow of OSOCE, like there is Spelling now, that we'll run in parallel as a test for now.
- [x] Contrast the number of false positives of `check-spelling` and codespell. How many inline ignores and dictionary entries? Also in the other projects.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-960)

Piedone:
[OSOE-960](https://lombiq.atlassian.net/browse/OSOE-960)
Fixes #423

[OSOE-960]: https://lombiq.atlassian.net/browse/OSOE-960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DemeSzabolcs (approved):

DemeSzabolcs:
The spell checking is good on `dev`. It was only triggered here because I switched to the `dev` branch in workflows, so it used the word library that is there, but since it's merged now, it includes "codespell" and "codespellrc".

PR #485, opened 2025-06-24
OSOE-1124: Add the ability to specify the Node.js version for builds

Piedone:
[OSOE-1124](https://lombiq.atlassian.net/browse/OSOE-1124)

[OSOE-1124]: https://lombiq.atlassian.net/browse/OSOE-1124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -1,70 +1,59 @@
 name: Set up Node.js
 description: >
-  Sets up Node.js with a specific version and exposes the `NODE_VERSION_TO_USE` environment variable with the specific
-  Node.js version that was set up. Intentionally not documented in Actions.md since it's only meant for internal use.
+  Runs actions/setup-node. Exists only to centralize which version of the action we use and exposes the
+  `NODE_VERSION_TO_USE` environment variable with the specific Node.js version that was set up. Intentionally not
+  documented in Actions.md since it's only meant for internal use.
 
+# Inputs are copied from https://github.com/actions/setup-node/blob/main/action.yml and formatted, as well as the
+# default Node.js version is specified.
 inputs:
+  always-auth:
+    description: 'Set always-auth in npmrc.'
+    default: 'false'
   node-version:
-    required: false
+    description: 'Version Spec of the version to use. Examples: 12.x, 10.15.1, >=10.15.0.'
     default: 22.16.0
-    description: >
-      Version Spec of the version to use. Examples: 12.x, 10.15.1, >=10.15.0.
-      See the documentation of https://github.com/actions/setup-node.
   node-version-file:
-    required: false
     description: >
       File containing the version Spec of the version to use. Examples: package.json, .nvmrc, .node-version,
       .tool-versions.
   architecture:
-    required: false
     description: >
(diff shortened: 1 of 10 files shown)


.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,97 @@
+name: Set up Node.js
+description: >
+  Sets up Node.js with a specific version exposes the `NODE_VERSION_TO_USE` environment variable with the specific
```
DemeSzabolcs:
```suggestion
  Sets up Node.js with a specific version and exposes the `NODE_VERSION_TO_USE` environment variable with the specific
```

or

```suggestion
  Sets up Node.js with a specific version, exposes the `NODE_VERSION_TO_USE` environment variable with the specific
```

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,97 @@
...
+    required: false
+    default: 22.16.0
+    description: >
+      Version Spec of the version to use. Examples: 12.x, 10.15.1, >=10.15.0.
+      See the documentation of https://github.com/actions/setup-node.
+  node-version-file:
+    required: false
+    description: >
+      File containing the version Spec of the version to use. Examples: package.json, .nvmrc, .node-version,
+      .tool-versions.
```
DemeSzabolcs: Since you can set both, I would also document which one takes precedence.

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,97 @@
...
+        cache-dependency-path: ${{ inputs.cache-dependency-path }}
+        always-auth: ${{ inputs.always-auth }}
+        mirror: ${{ inputs.mirror }}
+        mirror-token: ${{ inputs.mirror-token }}
+
+    - name: Save Node.js Version Into Environment Variable
+      shell: pwsh
+      run: |
+        $nodeVersion = node --version
+        Set-GitHubEnv 'NODE_VERSION_TO_USE' $nodeVersion
```
DemeSzabolcs: I was also thinking of printing this value out, so we can see it from the workflow run. However, the user can still figure it out, I guess: 
image

Do you think that's needed?

DemeSzabolcs (changes requested):

.github/actions/setup-node/action.yml
Piedone: Note that this is a blanket copy of the `actions/setup-node` inputs, the point being to have a centralized (and slightly extended) call to `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020`. We have other similar such internal actions, like `checkout`.

So, I'd keep this as-is, but I've documented this (and updated with the latest https://github.com/actions/setup-node/blob/main/action.yml).

.github/actions/setup-node/action.yml
Piedone: Yeah, that would be nice, adding it.

DemeSzabolcs (approved):

PR #494, opened 2025-07-25
OSOE-737: Add action to push (mirror) changes to a different repository

### LINKED ISSUE #285: Add action to push (mirror) changes to a different repository (OSOE-737)
For example to sync changes from our working repository of a project to release to the client's. Check if there are suitable actions in the [GitHub Marketplace](https://github.com/marketplace?category=&query=mirror+sort%3Apopularity-desc&type=actions&verification=) to wrap, e.g., https://github.com/marketplace/actions/mirror-repository.

Empty `git push`es must be prevented by checking the state of the target remote with `git log`, also see https://github.com/Lombiq/Git-hg-Mirror-Daemon/issues/2.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-737)

BenedekFarkas:
[OSOE-737](https://lombiq.atlassian.net/browse/OSOE-737)
Fixes #285

[OSOE-737]: https://lombiq.atlassian.net/browse/OSOE-737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Docs/Workflows/DeveloperTools/Codespell.md
Piedone, line 1: Don't move these, external links will break.

Docs/Workflows/GitHubAutomation/MirrorBranches.md
```diff
@@ -0,0 +1,70 @@
...
+    secrets:
+      SOURCE_TOKEN: ${{ secrets.LOMBIQ_CLIENT_PROJECT_LOMBIQ_TOKEN }}
+      DESTINATION_TOKEN: ${{ secrets.LOMBIQ_CLIENT_PROJECT_CLIENT_TOKEN }}
+```
+
+### 2. Periodically mirror branches from another repository
+
+Mirroring branches on a schedule and manual trigger with the workflow running in the destination repository.
+
+Note that this workflow doesn't necessarily need to be in either the source or destination repository; in that case the job's condition regarding the repository need to be omitted.
```
Piedone: Document what to do with branches if this workflow is in the repository that we mirror to. I guess the default branch in the destination repo (i.e. the one that needs to contain the version of this workflow used for the scheduled trigger) shouldn't have the same name as a branch in the source repo.

.github/workflows/mirror-branches.yml
Piedone, line 1: You did a lot of test runs. Please link one for each of the scenarios.

.github/workflows/mirror-branches.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ${{ inputs.machine-type }}
+    name: Mirror branches
+    if: inputs.destination-repository != inputs.source-repository
+    timeout-minutes: ${{ inputs.timeout-minutes }}
+    steps:
+      - name: Push to destination
+        uses: Lombiq/GitHub-Actions/.github/actions/mirror-branches@issue/OSOE-737
+        with:
+          source-repository: ${{ inputs.source-repository }}
+          source-token: ${{ secrets.SOURCE_TOKEN }}
```
Piedone, line 75: Pass the tokens as `env` instead of `with` parameters. That's a bit more secure against accidental disclose. Same for `get-branches`.

.github/actions/mirror-branches/action.yml
```diff
@@ -0,0 +1,182 @@
...
+
+        if (-not $destinationRepository -match '^[^/]+/[^/]+$')
+        {
+          throw 'Please provide a valid destination repository path.'
+        }
+
+        if ($sourceRepository.ToLower() -eq $destinationRepository.ToLower())
+        {
+          throw 'The source and the destination repositories cannot be the same.'
+        }
```
Piedone:
```suggestion
        if (-not $branchNames -and -not $branchRegex)
        {
            throw 'At least one of the ''branch-names'' and ''branch-regex'' parameters must be non-empty.'
        }

        if (-not $sourceRepository -match '^[^/]+/[^/]+$')
        {
            throw 'Please provide a valid source repository path.'
        }

        if (-not $destinationRepository -match '^[^/]+/[^/]+$')
        {
            throw 'Please provide a valid destination repository path.'
        }

        if ($sourceRepository.ToLower() -eq $destinationRepository.ToLower())
        {
            throw 'The source and the destination repositories cannot be the same.'
        }
```

.github/actions/get-branches/action.yml
```diff
@@ -0,0 +1,58 @@
...
+
+        if ($allBranches)
+        {
+          Write-Output "::notice::Found $($allBranches.Count) branches in '${{ inputs.repository }}'."
+          Set-GitHubOutput 'branches' ($allBranches | ConvertTo-Json -Depth 9 -Compress)
+        }
+        else
+        {
+          Write-Output '::warning::''${{ inputs.repository }}'' has no branches!'
+        }
```
Piedone:
```suggestion
        if ($allBranches)
        {
            Write-Output "::notice::Found $($allBranches.Count) branches in '${{ inputs.repository }}'."
            Set-GitHubOutput 'branches' ($allBranches | ConvertTo-Json -Depth 9 -Compress)
        }
        else
        {
            Write-Output '::warning::''${{ inputs.repository }}'' has no branches!'
        }
```

.github/actions/get-branches/action.yml
Piedone, line 1: Does this need to be an action instead of a script in the `mirror-branches`' folder?

.github/actions/mirror-branches/action.yml
```diff
@@ -0,0 +1,182 @@
...
+    - name: Find branches to mirror
+      id: find-branches
+      shell: pwsh
+      run: |
+        $branchNames = '${{ inputs.branch-names }}'.Trim()
+        $branchRegex = '${{ inputs.branch-regex }}'.Trim()
+
+        if ($branchNames)
+        {
+          $branchNameList = $branchNames -split ',' | ForEach-Object { $PSItem.Trim() } | Where-Object { $PSItem -ne '' }
```
Piedone: The `if`s have two spaces as indentation throughout this step.

.github/actions/mirror-branches/action.yml
```diff
@@ -0,0 +1,182 @@
...
+      id: get-destination-branches
+      uses: Lombiq/GitHub-Actions/.github/actions/get-branches@issue/OSOE-737
+      with:
+        repository: ${{ inputs.destination-repository }}
+        api-token: ${{ inputs.destination-token }}
+
+    - name: Find branches to mirror
+      id: find-branches
+      shell: pwsh
+      run: |
```
Piedone, line 90: This is quite long and would be more appropriate in a separate ps1 file, where it also gets code analysis.

.github/actions/mirror-branches/action.yml
```diff
@@ -0,0 +1,182 @@
...
+        Set-GitHubOutput 'branch-names' ($mirroringBranchNames -join ',')
+
+    - name: Checkout source repository
+      if: steps.find-branches.outputs.first-branch-name != ''
+      uses: Lombiq/GitHub-Actions/.github/actions/checkout@dev
+      with:
+        repository: ${{ inputs.source-repository }}
+        ref: ${{ steps.find-branches.outputs.first-branch-name }}
+        token: ${{ inputs.source-token }}
+        fetch-depth: 0
```
Piedone, line 108: `fetch-depth: 0` fetches the whole repo, all branches and all history. Did you mean `fetch-depth: 1`? But then that branch needs to be fully fetched too in the next step.

.github/actions/mirror-branches/action.yml
```diff
@@ -0,0 +1,182 @@
...
+        api-token: ${{ inputs.source-token == '' && inputs.destination-token || inputs.source-token }}
+
+    - name: Get destination branches
+      id: get-destination-branches
+      uses: Lombiq/GitHub-Actions/.github/actions/get-branches@issue/OSOE-737
+      with:
+        repository: ${{ inputs.destination-repository }}
+        api-token: ${{ inputs.destination-token }}
+
+    - name: Find branches to mirror
```
Piedone, line 87: Do I understand correctly that to prevent empty pushes, the logic here is to poll the branches of the source repo and compare if their head commits equal to the head commits of the corresponding destination repo branches?

Piedone (changes requested):

Docs/Workflows/DeveloperTools/Codespell.md
BenedekFarkas: Reverted.

Docs/Workflows/GitHubAutomation/MirrorBranches.md
BenedekFarkas: Are you referring to the case where for example our private mirror of https://github.com/Lombiq/DotNest-Core-SDK has a different default branch (`dev-site`)?

.github/workflows/mirror-branches.yml
BenedekFarkas: 1. Mirror on push: https://github.com/Lombiq/OSOE-737-Test-1/actions/runs/16333035463
2. Mirror on schedule from another repo: https://github.com/Lombiq/OSOE-737-Test-1/actions/runs/16528417935
3. Mirror on schedule from a repo to another one running in a third repo: https://github.com/Lombiq/OSOE-737-Test-3/actions/runs/16605865746

.github/actions/get-branches/action.yml
BenedekFarkas: Modified your suggestion by fixing syntax (`${{ inputs.repository }}`).

.github/actions/get-branches/action.yml
BenedekFarkas: It doesn't need to be, but the functionality is generic and gives LGHA more "surface area" for 3rd-party usage. On that note I forgot to add it to `Actions.md`.

Docs/Workflows/GitHubAutomation/MirrorBranches.md
Piedone: Yep.

.github/workflows/mirror-branches.yml
```diff
@@ -0,0 +1,75 @@
...
+        default: .+
+
+jobs:
+  mirror:
+    runs-on: ${{ inputs.machine-type }}
+    name: Mirror branches
+    if: inputs.destination-repository != inputs.source-repository
+    timeout-minutes: ${{ inputs.timeout-minutes }}
+    steps:
+      - name: Push to destination
```
Piedone: Since this is doing all the mirroring:

```suggestion
      - name: Mirror branches
```

.github/workflows/mirror-branches.yml
Piedone: Thanks!

.github/workflows/mirror-branches.yml
BenedekFarkas: Ah yes - it wasn't some iterations ago.

.github/actions/mirror-branches/action.yml
BenedekFarkas: Correct. `git log` works too as discussed earlier, but for that the destination branches need to be fetched too and then checked individually.

.github/actions/mirror-branches/action.yml
BenedekFarkas: To be fixed in the factored out script.

.github/actions/mirror-branches/action.yml
Piedone: Great, thanks!

.github/actions/mirror-branches/action.yml
BenedekFarkas: Based on an earlier iteration, I concluded that if you provide a `ref` when using `fetch-depth: 0` then it only applies to the ref specified, because I still had to fetch all the other branches manually. I'll check again to be sure.

Piedone: Ah, I totally missed the `ref` being supplied, then nevermind.

Docs/Workflows/GitHubAutomation/MirrorBranches.md
BenedekFarkas: OK, here it is: 94987730f0366f5635b5ee263a123a2fb880817b

.github/workflows/mirror-branches.yml
BenedekFarkas: I did that in an earlier iteration, but also found that we have a mixed usage of `env` and `with` when it comes to secrets. Regular parameters have the added benefit of the `required` setting and secrets are always obfuscated in the output logs regardless of they are passed into the action/workflow.

Piedone (approved):

.github/actions/mirror-branches/action.yml
Piedone, line 1: It just occurred to me, that it's important that we know about these workflows failing via some notification. However, by default, only that user will be notified about a scheduled workflow's failure who last modified the file. This isn't necessarily good enough. So, we need something that can reach multiple (configurable) people.

With any text field on GitHub this is trivial, since you can just @mention teams too (this is what we do e.g. in the PR descriptions of OSOCE Renovate updates, like https://github.com/Lombiq/UI-Testing-Toolbox/pull/577). However here I don't think we can just easily comment. But perhaps on failure the action could comment on the last commit or something.

Alternatively, it could send a Teams notification (there are actions for this).

BenedekFarkas: > It just occurred to me, that it's important that we know about these workflows failing via some notification.

Indeed true for the ones scheduled by a trigger. Not necessary for the ones triggered manually or by push.

> But perhaps on failure the action could comment on the last commit or something.

We can mention a specific user/group who should take care of that, but the author of that commit would get notified too, even though the commit and its author might have nothing do with the failure.

> Alternatively, it could send a Teams notification (there are actions for this).

Teams is better IMO, although some of the workflows run on a schedule, so it can spam a lot of messages before anyone notices.

-----

How about creating a GitHub issue and a Jira issue using the `create-jira-issues-for-community-activities` workflow? It should use (and look for) a specific title pattern and not create an issue if there's already an open issue with it.
All this being optional, so rather than being part of the `mirror-branch` action/workflow, it can be demonstrated in `MirrorBranches.md`.

Piedone: Yeah, perhaps this is only necessary for scheduled workflows.

Creating jut a GitHub issue can be enough, since you can subscribe to new issues on a repo (and then we don't need to maintain an Atlassian API token for those repos to create Jira issues). Although that would be an exception to our usual workflow of not having GitHub issues even enabled in private projects.

Hmm, in the end, just Teams notifications would be enough, I think. Those can simply call a permanent webhook. And while a notification flood is annoying, it should be really rare.

BenedekFarkas: > Creating jut a GitHub issue can be enough, since you can subscribe to new issues on a repo (and then we don't need to maintain an Atlassian API token for those repos to create Jira issues). Although that would be an exception to our usual workflow of not having GitHub issues even enabled in private projects.

Agreed.

> Hmm, in the end, just Teams notifications would be enough, I think. Those can simply call a permanent webhook. And while a notification flood is annoying, it should be really rare.

Famous last words. :D
But yeah, I hope it will remain fairly stable and the Teams message webhook is a simple fire-and-forget solution. Let's start with that.

Piedone: Well, yeah :). We can do a sparse schedule for most of the workflows so even if one fails and starts sending continuous notifications it won't be a huge annoyance.

Piedone (approved):

PR #509, opened 2025-09-17
OLYS-34: Add package-manager-cache option to build-and-test-orchard-core.yml

sarahelsaig:
[OLYS-34](https://lombiq.atlassian.net/browse/OLYS-34)

[OLYS-34]: https://lombiq.atlassian.net/browse/OLYS-34?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -44,6 +44,11 @@ inputs:
     description: >
       Used to specify a package manager for caching in the default directory. Supported values: npm, yarn,
       pnpm.
+  package-manager-cache:
+    description: >
+      Set to false to disable automatic caching based on the package manager field in package.json. By default,
+      caching is enabled if the package manager field is present.
+    default: true
   cache-dependency-path:
     description: >
       Used to specify the path to a dependency file: package-lock.json, yarn.lock, etc. Supports wildcards or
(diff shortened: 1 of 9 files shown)


.github/actions/setup-node/action.yml
```diff
@@ -66,6 +71,7 @@ runs:
     - name: Set up Node.js
       uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
       with:
+        always-auth: ${{ inputs.always-auth }}
```
barthamark, line 74: Shouldn't you give it a description and default value like for the `package-manager-cache`?

barthamark (changes requested):

.github/actions/setup-node/action.yml
sarahelsaig: This was already here. I only moved it to the start of the list so the items in the `with:` block match the order of the `inputs:` above. Also I have no idea what it actually does and the description in the inputs [is copied from upstream](https://github.com/actions/setup-node/blob/main/action.yml#L6) so I don't want to touch it.

barthamark: Ah yes I didn't see it was already here just moved it.

PR #522, opened 2025-10-22
OSOE-1010: Updating default Node version to 24.10.0, removing PNPM workaroun

Piedone:
[OSOE-1010](https://lombiq.atlassian.net/browse/OSOE-1010)

[OSOE-1010]: https://lombiq.atlassian.net/browse/OSOE-1010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -47,6 +47,12 @@ For .NET workflows, the default .NET SDK version we should provide is a concrete
 
 We can still choose to update to a more recent patch version, but only deliberately.
 
+### Default Node.js version
+
+.NET build workflows install the latest Active Long-Term Support (LTS) version of Node.js by default, but the version is configurable. You can check out the [Node.js release schedule](https://nodejs.org/en/about/previous-releases) to see which version is the latest Active LTS at any given time.
+
+Build workflows also enable [corepack](https://github.com/nodejs/corepack) to allow using Node.js package managers other than NPM. This behavior will be disabled when upgrading to Node.js v26 since that version [won't contain corepack anymore](https://github.com/nodejs/node/issues/50963#issuecomment-2746471823).
+
 ### Reference validation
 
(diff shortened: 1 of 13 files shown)


Readme.md
```diff
@@ -47,6 +47,12 @@ For .NET workflows, the default .NET SDK version we should provide is a concrete
 
 We can still choose to update to a more recent patch version, but only deliberately.
 
+### Default Node.js version
+
+.NET build workflows install the latest Active Long-Term Support (LTS) version of Node.js by default, but the version is configurable. You can check out the [Node.js release schedule](https://nodejs.org/en/about/previous-releases) to see which version is the latest Active LTS at any given time.
+
+Build workflows also enable [corepack](https://github.com/nodejs/corepack) to allow using Node.js package managers other than NPM. This behavior will be disabled when upgrading to Node.js v26 since that version [won't contain corepack anymore](https://github.com/nodejs/node/issues/50963#issuecomment-2746471823).
```
sarahelsaig, line 54: Is this warning line necessary? It seems like after that point we can just replace `corepack enable` with `corepack enable to npm i -g corepack` so the users don't necessarily need to know about this. Or if you want to be fancy:

```pwsh
$nodeVersion = [version] $(node -v).Substring(1)

if ($nodeVersion.Major -ge 25)
{
    npm install --global corepack
}
else
{
    corepack enable
}
```

sarahelsaig (changes requested):

Readme.md
Piedone: That's clever, but I'd stop supporting corepack. At that point, it'll be quasi-deprecated, and we certainly won't use it. So I'd rather nudge everybody to not use it (or stick to v24).

sarahelsaig: Okay then.

sarahelsaig (approved):

PR #534, opened 2025-11-17
OSOE-1194: Freeing up runner storage space before .NET builds

Piedone:
[OSOE-1194](https://lombiq.atlassian.net/browse/OSOE-1194)

[OSOE-1194]: https://lombiq.atlassian.net/browse/OSOE-1194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
+++ b/.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
@@ -7,7 +7,7 @@ df --human-readable
 sudo apt-get clean
 
 # Remove Android SDK.
-sudo rm -rf /usr/local/lib/android
+sudo Remove-Item -Recurse -Force /usr/local/lib/android
 
 # We can't remove all unused Docker images since that would also remove the Elasticsearch image (used in the Set up
 # Elasticsearch step) and any possible Docker images from prepared runner images.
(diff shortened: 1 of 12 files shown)


.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
```diff
@@ -0,0 +1,42 @@
+$initialSpace = (df / |
+        Select-Object -Skip 1 |
+        ForEach-Object { $PSItem.Split(' ', [System.StringSplitOptions]::RemoveEmptyEntries)[3] } |
+        Measure-Object -Sum).Sum
```
sarahelsaig: Why not use .NET instead of parsing a utility? It should be more reliable. You can get the "drive" for `/` like this:

```pwsh
$rootDrive = (Get-PSDrive /)[0]
$initialSpace = $rootDrive.Free # Free is a dynamic property so it can be reused below.
```

.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+df --human-readable
+
+# Remove the local cache of downloaded packages but don't remove installed software.
+sudo apt-get clean
+
+# Remove Android SDK.
+sudo rm -rf /usr/local/lib/android
```
sarahelsaig, line 13: Is there a particular reason for using `rm -rf` instead of `Remove-Item -Recurse -Force`? If so, please include a comment.

sarahelsaig (changes requested):

.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
Piedone: No reason, actually, other than me starting with Linux only, and the action then gradually evolving.

.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
Piedone: No good reason either, just historical. Thanks for the code.

.github/actions/free-up-space-for-dotnet/FreeUpSpaceUbuntu.ps1
Piedone: `rm -rf` is better it turns out, commented.

sarahelsaig (approved):

PR #559, opened 2026-01-04
OSOE-1101: Indicate if a PR has breaking changes and generate CompatibilitySuppressions.xml file files

### LINKED ISSUE #466: Indicate if a PR has breaking changes, as discovered by `validate-nuget-publish` (OSOE-1101)
`validate-nuget-publish` (and `publish-nuget`) fails if the NuGet baseline package validation finds breaking changes. Sometimes these are not actual breaking changes (like constructor injections changing), but mostly are, and we're either consciously suppressing these errors if breaking changes are expected, or need to avoid introducing the breaking change.

For the expected case, we could have an indication in the PR, foremost in its title and with a label, that it contains breaking changes. That way, it's more apparent to the reviewer and more visible in the automatically generated release notes on GitHub.

- [x] Have an optional feature usable via both `validate-nuget-publish` and `publish-nuget` that, when baseline package validation fails, adds the "(⚠️ breaking changes)" text at the end of the PR title (if a PR exists; for `publish-nuget` this is not necessarily the case), and the `breaking-changes` label. Additionally, it should also add a comment under the PR if one exists (if https://github.com/Lombiq/GitHub-Actions/issues/467 was done earlier, change the comment to refer to the available _CompatibilitySuppressions.xml_ file):
```
This pull request appears to contain breaking changes.
- If you can do this without breaking changes, then please do that, it'd be better.
- If breaking changes are inevitable, then add an explanation of these and a migration guide for users to the description of the pull request (during the next release, these will be added to the release notes). Generate the _CompatibilitySuppressions.xml_ file following [the corresponding docs](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions?tab=readme-ov-file#opening-pull-requests).
- If the changes are technically breaking but not something that would actually be breaking for consuming projects (like the signature of a constructor changing due to new injected dependencies) then also just generate the _CompatibilitySuppressions.xml_ file. Also consider how in the future this can be avoided: Can perhaps the class be made `internal` (what's usually better for implementations of interfaces for injected dependencies)?
```
- [x] If package validation passes without any new suppressions (prior suppressions might be there), the title suffix and the label should be removed if they exist.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-1101)

### LINKED ISSUE #467: If NuGet baseline package validation fails, generate a new artifact with the new CompatibilitySuppressions.xml file (OSOE-1103)
The NuGet workflows have all the information ready to prepare this file for us, so it would be nice.

- [x] Generate a _CompatibilitySuppressions.xml_ file, both in `validate-nuget-publish` and `publish-nuget`, if NuGet baseline package validation fails, and make it downloadable as an artifact. This needs `/p:GenerateCompatibilitySuppressionFile=true` in the `dotnet pack` calls.
- [x] Update the docs under https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions?tab=readme-ov-file#opening-pull-requests.
- [x] If https://github.com/Lombiq/GitHub-Actions/issues/466 was done earlier, then change the comment added by it to refer to using this artifact instead of instructing how to generate the file.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-1103)

Piedone:
[OSOE-1101](https://lombiq.atlassian.net/browse/OSOE-1101)
Fixes #467
Fixes #466

[OSOE-1101]: https://lombiq.atlassian.net/browse/OSOE-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/mark-breaking-changes/Update-PullRequestTitle.ps1
```diff
@@ -0,0 +1,30 @@
...
+    {
+        $newTitle = $currentTitle + $suffix
+    }
+}
+else
+{
+    if ($currentTitle.Contains($suffix))
+    {
+        $newTitle = $currentTitle.Substring(0, $currentTitle.Length - $suffix.Length)
+    }
```
sarahelsaig, line 24: If for whatever reason the `$suffix` is not the very end of the string (i.e. if any other script adds more strings at the end), this will break. Either replace `.Contains()` with `.EndsWith()`, or use `.Replace()` instead of `.Substring()`. I think the latter is more flexible.

.github/actions/mark-breaking-changes/Update-PullRequestTitle.ps1
sarahelsaig, line 1: How about moving this file to _/Scripts_, adding `$Suffix` as a parameter and renaming `$IsBreaking` to `$Remove`, so this can be reused for other tasks in the future?

.github/actions/mark-breaking-changes/action.yml
```diff
@@ -0,0 +1,63 @@
+name: Mark Breaking Changes
+description: >
+  Adds or removes a breaking-changes indication on a pull request title and label, optionally leaving a comment.
+  Intentionally not documented in Actions.md since it's only meant for internal use.
```
sarahelsaig, line 4: Why is it necessary to both alter title and add label? I'd think just the label is enough, like it happens with `Merge Conflict`.

sarahelsaig (changes requested):

.github/actions/mark-breaking-changes/action.yml
Piedone: This is so it's apparent from a generated release notes which PRs are breaking. There the PRs are listed with their titles, e.g. https://github.com/Lombiq/GitHub-Actions/releases/tag/v4.1.0. So, when editing that, it's easier to see breaking changes there than having to separately list them by label.

.github/actions/mark-breaking-changes/Update-PullRequestTitle.ps1
Piedone: I'd rather do that if and when that happens, because then we'll know how best to refactor. There's very little logic here, and `Update-GitHubIssueAndPullRequest.ps1`, the only other place where we change the PR title, wouldn't be able to use this.

sarahelsaig: > I'd rather do that if and when that happens, because then we'll know how best to refactor. There's very little logic here

Sure, we can leave that for later but these feel like contradictory sentiments to me. 🤷‍♀️

> Update-GitHubIssueAndPullRequest.ps1, the only other place where we change the PR title

Now that you've brought that script to my attention, I notice that the parameter over there is `-PullRequestId` while here it's `-PullRequestNumber`. Can we bring them into harmony, so that there is no confusion that they refer to the same thing?

.github/actions/mark-breaking-changes/action.yml
sarahelsaig: I see, that makes sense.

.github/actions/mark-breaking-changes/Update-PullRequestTitle.ps1
Piedone: Sure. The correct term (following what GitHub uses in the GHA context) is "number".

sarahelsaig (approved):

PR #598, opened 2026-04-20
OSOE-1167: Optionally fail the build workflow if the repository changed after the .NET build

### LINKED ISSUE #518: Optionally fail the build workflow if the repository changed after the .NET build (OSOE-1167)
A repository and the build pipeline in it should be set up in a way that building the projects in it won't cause pending changes in Git. So, add a feature to optionally enable a check for this, and fail the workflow if there are pending changes, in the `build-and-test-*` and `msbuild-and-test` workflows.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-1167)

wAsnk:
[OSOE-1167](https://lombiq.atlassian.net/browse/OSOE-1167)
Fixes #518

[OSOE-1167]: https://lombiq.atlassian.net/browse/OSOE-1167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/verify-git-status-clean/action.yml
+++ b/.github/actions/verify-git-status-clean/action.yml
@@ -1,0 +1,57 @@
+name: Verify Git Status is Clean
+description: Fails if the current Git repository has pending tracked or untracked changes.
+
+inputs:
+  ignored-paths:
+    required: false
+    default: ''
+    description: >
+      Newline-separated repository-relative paths to ignore when checking for pending Git changes.
+
+runs:
+  using: composite
+  steps:
+    - name: Verify Git Status is Clean
+      shell: pwsh
+      run: |
+        $ignoredPathsInput = @'
+        ${{ inputs.ignored-paths }}
+        '@
+
+        $ignoredPaths = @()
+        if (-not [string]::IsNullOrWhiteSpace($ignoredPathsInput))
+        {
+            $ignoredPaths = $ignoredPathsInput -split "`r?`n" |
+              ForEach-Object { $PSItem.Trim() } |
+              Where-Object { $PSItem }
+        }
+
+        $pendingChanges = @(git status --porcelain=v1 --untracked-files=all)
+
+        if ($LASTEXITCODE -ne 0)
+        {
+            Write-Output '::error::Failed to get Git status.'
+            exit $LASTEXITCODE
+        }
+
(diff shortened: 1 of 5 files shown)


.github/actions/verify-git-status-clean/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        {
+            $path = $pendingChange.Length -gt 3 ? $pendingChange.Substring(3) : $pendingChange
+
+            if ($path -in $ignoredPaths)
+            {
+                continue
+            }
+
+            $relevantChanges += $pendingChange
+        }
```
sarahelsaig: So you mean like this?
```pwsh
$relevantChanges = $pendingChanges |
    Where-Object { ($PSItem.Length -gt 3 ? $PSItem.Substring(3) : $PSItem) -notin $ignoredPaths}
```

.github/actions/verify-git-status-clean/action.yml
sarahelsaig, line 1: Can you add a final step where the [`upload-artifact` action](https://github.com/Lombiq/GitHub-Actions/blob/76e0fc233e82012e34b346a2c118119fa5c021f8/.github/actions/upload-artifact/action.yml) is called on the files that cause a run failure? Note that [according to this](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions) the `path` can be multiple files separated by newlines, so you can output `$relevantChanges` from the previous step.

sarahelsaig (changes requested):

.github/actions/verify-git-status-clean/action.yml
wAsnk: I'd ask @Piedone if we want to add this. Let me know if I should add this to the scope.

.github/actions/verify-git-status-clean/action.yml
wAsnk: Yeah, that is better! Thanks, updated this part!

.github/actions/verify-git-status-clean/action.yml
Piedone: Yes.

sarahelsaig (approved):

PR #602, opened 2026-04-25
OSOE-1199: Migrate solution files to SLNX

sarahelsaig:
[OSOE-1199](https://lombiq.atlassian.net/browse/OSOE-1199)

[OSOE-1199]: https://lombiq.atlassian.net/browse/OSOE-1199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahelsaig:
Ignore the validation errors here, the checker has trouble when the target of a PR is not `dev`.

.github/workflows/validate-this-gha-refs.yml
DemeSzabolcs: I think this should be reverted, if the validation fails anyway, no?

.github/actions/publish-nuget/New-NuGetPackage.ps1
```diff
@@ -84,7 +84,7 @@ if ($doBaselinePackageValidation -and -not (Test-Path -Path $CompatibilitySuppre
 # Whether the pull request contains breaking changes is a global, not project-specific property.
 $isBreaking = $false
 
-$projects = (Test-Path *.sln) ? (dotnet sln list | Select-Object -Skip 2 | Get-Item) : (Get-ChildItem *.csproj)
+$projects = ((Test-Path *.slnx) -or (Test-Path *.sln)) ? (dotnet sln list | Select-Object -Skip 2 | Get-Item) : (Get-ChildItem *.csproj)
```
DemeSzabolcs, line 87: Can't we also use the wild card here? `*.sln?"`.

DemeSzabolcs (changes requested):

.github/actions/publish-nuget/New-NuGetPackage.ps1
sarahelsaig: No. That works only for `Get-ChildItem -Filter *.sln?`, because its `-Filter` argument [uses a special non-standard grep syntax](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.6#-filter).

For example `Get-ChildItem *.sln?` doesn't work either. That's because for `-Filter` the `?` means zero or one character, while in normal Powershell grep syntax `?` means exactly one character.

DemeSzabolcs: Okay!

DemeSzabolcs (approved):

.github/workflows/build-and-test-orchard-core.yml
```diff
@@ -367,7 +367,7 @@ jobs:
 
       - name: Tests
         if: inputs.test-disable == 'false'
-        uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@dev
+        uses: Lombiq/GitHub-Actions/.github/actions/test-dotnet@issue/OSOE-1199
```
DemeSzabolcs: This stayed `dev` in some places. Like in [build-and-test-dotnet.yml](https://github.com/Lombiq/GitHub-Actions/pull/602/changes#diff-4d27efb3431c203470e1312efe596962845c94cda34d5a9876947a9098c1e798)

So it should be consistent.

.github/workflows/msbuild-and-test.yml
DemeSzabolcs, line 226: This is a new action; shouldn't it point to `OSOE-1199`? Because it won't be merged to `dev` first, but to `OSOE-1199` .

DemeSzabolcs (changes requested):

.github/workflows/msbuild-and-test.yml
sarahelsaig: It's not actually a new action, [it's already in dev](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/verify-git-status-clean/action.yml). I think GH is just glitching. :/

DemeSzabolcs: Ah okay!

.github/workflows/build-and-test-orchard-core.yml
sarahelsaig: I've only changed what's necessary for building OSOCE. I think consistency doesn't matter here, because this is temporary. It will all go back to `dev` in the end.

sarahelsaig:
I've updated the refs so it's all green now.

DemeSzabolcs (approved):

PR #604, opened 2026-04-29
OSOE-1149: Only allow publishing a major release via `publish-nuget` if CompatibilitySuppressions files are present in Lombiq.GitHub.Actions

### LINKED ISSUE #500: Only allow publishing a major release via `publish-nuget` if CompatibilitySuppressions files are present (OSOE-1149)
If _CompatibilitySuppressions.xml_ files are present in the project, then that means that there are breaking changes. In such a case, the `publish-nuget` workflow (action) should only allow publishing a major release and fail otherwise.

- [x] Check for any _CompatibilitySuppressions.xml_ files in the repo. Keep in mind that a repo can include multiple projects, but since these are all published together, a _CompatibilitySuppressions.xml_ file (and thus breaking changes) in any warrants a major release. We don't care about the case of there being such a file in the repo without it actually meant for NuGet publishing (like as an example).
- [x] If a _CompatibilitySuppressions.xml_ file exists, then the version number should be a new major one above the current version. E.g. if the current version is 1.2.3, the new version should be e.g. 2.0.0. With version numbers being in the x.y.z format, what needs to be checked is that xNew > xCurrent, enforcing anything else is not necessary (so, e.g. after a 1.2.3 you can publish 2.3.4 or 3.0.0 too, we shouldn't prevent that). If there's no _CompatibilitySuppressions.xml_ file, then no check should be performed.
- [x] If an incompatible version is attempted to be published, fail the action with an error message explaining that following Semantic Versioning, you should publish a new major version.
- [x] Action and workflow input parameter to enable this, enabled being the default. This will cover that you decide in general if you want this, and also if you want to disable the check just for a given publish (temporarily, then reverting it). E.g. https://github.com/Lombiq/UI-Testing-Toolbox/pull/589 is technically a breaking change but not something that would actually be breaking for users. It would've been nicer to keep the `FakeStore` class internal originally.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-1149)

wAsnk:
[OSOE-1149](https://lombiq.atlassian.net/browse/OSOE-1149)
Fixes #500

Expected fail: https://github.com/Lombiq/Hosting-Media-Theme/actions/runs/25491487726/job/74800367839
Expected pass with override: https://github.com/Lombiq/Hosting-Media-Theme/actions/runs/25491668856/job/74800976068
Both in test PR: https://github.com/Lombiq/Hosting-Media-Theme/pull/109

[OSOE-1149]: https://lombiq.atlassian.net/browse/OSOE-1149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -141,7 +141,8 @@ runs:
 
         if ('${{ inputs.require-major-version-for-compatibility-suppressions }}' -eq 'true' -and -not $isPreReleaseVersion)
         {
-            $compatibilitySuppressionsFiles = @(Get-ChildItem -Path . -Recurse -File -Filter CompatibilitySuppressions.xml -ErrorAction SilentlyContinue)
+            $compatibilitySuppressionsFiles = @(
+                Get-ChildItem -Path . -Recurse -File -Filter CompatibilitySuppressions.xml -ErrorAction SilentlyContinue)
 
             if ($compatibilitySuppressionsFiles.Count -gt 0)
             {
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -152,17 +152,7 @@ runs:
                 Write-Output 'CompatibilitySuppressions.xml file(s) found:'
                 $compatibilitySuppressionsFiles | ForEach-Object { Write-Output "- $($PSItem.FullName)" }
 
-                $versionMatch = [regex]::Match($version, '^(?<major>\d+)\.\d+\.\d+($|-)')
-                if (-not $versionMatch.Success)
-                {
-                    Write-Output ("::error::CompatibilitySuppressions.xml files indicate breaking changes, but " +
-                        "`"$version`" doesn't use an x.y.z version format. Following Semantic Versioning, publish a " +
-                        'new major version.')
-
-                    exit 1
-                }
-
-                $publishMajorVersion = [int]$versionMatch.Groups['major'].Value
+                $publishMajorVersion = ([Version]$version).Major
 
                 if ($publishMajorVersion -le $current.Major)
                 {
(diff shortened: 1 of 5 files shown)


.github/actions/publish-nuget/action.yml
```diff
@@ -129,6 +137,42 @@ runs:
             $baselineVersion = ''
         }
 
+        $isPreReleaseVersion = $version -match '^\d+\.\d+\.\d+-(alpha|beta|preview|rc)(\.\d+(\.[a-zA-Z]{2,}(-\d+)?)?)?$'
```
sarahelsaig: There is already a version format regex in line 98 and if it ever changes that could be a pretty nasty future error source. Make a variable for the regex (e.g. `$versionRegex`) and reuse it here:
```pwsh
$version -match $versionRegex
$isPreReleaseVersion = $Matches.Values.Count -gt 1
```
(`$Matches` is a magic variable that Powershell populates with the Regex match groups whenever you use the `-match` operator.)

.github/actions/publish-nuget/action.yml
```diff
@@ -129,6 +137,42 @@ runs:
...
+                if (-not $versionMatch.Success)
+                {
+                    Write-Output ("::error::CompatibilitySuppressions.xml files indicate breaking changes, but " +
+                        "`"$version`" doesn't use an x.y.z version format. Following Semantic Versioning, publish a " +
+                        'new major version.')
+
+                    exit 1
+                }
+
+                $publishMajorVersion = [int]$versionMatch.Groups['major'].Value
```
sarahelsaig: We can only reach to this point if the version number is normal semantic version, so the check above shouldn't really be necessary. More importantly, this whole operation can be simplified if you cast the string into `Version` object. (where casting will throw a pretty clear `Cannot convert value "..." to type "System.Version".` error if this somehow isn't true)


```suggestion
                $publishMajorVersion = ([Version]$version).Major
```

sarahelsaig (changes requested):

sarahelsaig (approved):

